PR #1, opened 2015-04-16
Update index.md

seneubert:
Added 3 topics

### DIFF
--- a/index.md
+++ b/index.md
@@ -15,12 +15,15 @@ Paragraph(s) of introductory material.
 
 This is a rough list of the areas we would like to cover:
 
-* first DecayTreeTuple on the grid (including screen, EOS storage, ...)
+* Setting up your development environment
+* Working with the repository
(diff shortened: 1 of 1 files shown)


betatim:
We need some stats to inform us how many of the absolute beginners (that just learned about bash and python the day before) will be doing development stuff. Think we should really try and keep it as minimal as possible and push all the nice (and interesting) things into `lhcb-second-analysis-steps`.

I remain shocked at how few people ever run Brunel themselves or even how few students make their own nTuples. From which I extrapolate that development is a long way up the road from first steps :-/

betatim:
Note: I will run `make preview` just before merging the PR, makes it easier to see what is actually proposed as change. @betatim don't forget!

betatim:
I'll merge this but I am still not sure we should start with "developer" stuff. I think that it adds confusion (if you aren't a 'coder') and isn't needed to do the first few things. If we do get to LoKi functors and how to modify one/write your own then we need it.

PR #2, opened 2015-04-30
Added topic page on creating plots from ntuples in ROOT

seneubert:
This establishes the document and link.
Content will be added


betatim:
Closing this. After the switch to auto building of the html we can create a new PR for this that aims at `master` as target not `gh-pages`.

PR #18, opened 2015-05-13
Add lesson on running a stripping line on MC

### LINKED ISSUE #15: Switch from `strip20` to `strip21` in minimal DV example
After #7 we then have "woah, we used the old stripping, let's fix that". This is what this section is about. Should be done before #9

ibab:
This is my draft for the lesson on re-running a stripping line.
I'm not quite happy with the format.
Maybe a complete example should come first, followed by the explanations?
This fixes #15.


07-rerun-stripping.md
```diff
@@ -0,0 +1,133 @@
...
+    for line in stream.lines:
+        if line.name() == custom_line:
+            custom_stream.appendLines([line])
+
+stream = 'CustomStream'
+line = 'D2hhCompleteEventPromptDst2D2RSLine'
+
+dtt = DecayTreeTuple('TupleDstToD0pi_D0ToKpi')
+dtt.Inputs = [
+    '/Event/{0}/Phys/{1}/Particles'.format(stream, line)
```
alexpearce: Have you tried running this options file? I have a feeling that you don't need the `/Event/STREAM` prefix (because the re-run stripping doesn't go to usual stream location I think).

ibab:
Sorry, forgot to mention that I haven't tested it yet!
And you're right, it should just be `/Event/Phys/...`.

alexpearce:
I think it's just `Phys/{0}/Particles'.format(line)`.

ibab:
Mh, I thought it would just add the `/Event` in front if it's missing, so both should work.
I'm not getting any particles from that location, though. (Using both `/Event/Phys/...` and `Phys/...`)
I'll make this run later as soon as I have some time.

07-rerun-stripping.md
```diff
@@ -0,0 +1,133 @@
...
+~~~
+
+Run the stripping as part of our DaVinci job.
+Make sure to access `.sequence()` here.
+
+~~~ {.python}
+DaVinci().appendToMainSequence([sc.sequence()])
+~~~
+
+This is how your option file could look like now:
```
betatim: could -> should

(could is like koennte)

betatim:
Can we move the complete options file to `code/` and then include it in the `.md`? Would be nice if it is easy to do so people can use the options file directly without having to copy and paste it.

ibab:
I used "could" because the file might look different depending on how/where the snippets were added.
But I see how this might be confusing, I'll change it to "should".
Having a `code/` directory is a good idea.

ibab:
I've based the lesson on a new version of the minimal DV example and also reduced the amount of text. I'm thinking of pointing out the key things that have been added to rerun the stripping instead of going through the file line by line.

ibab:
I've updated the lesson. I would be okay with merging now.
The example seems to be working.
I'm just a bit concerned about the fact that I had to write the HltDecReport to a different location. (It was complaining that there was already a report there).
Wouldn't this be a problem if TupleToolStripping reads from it?
Maybe someone with more stripping knowledge can help out here?

14-rerun-stripping.md
```diff
@@ -0,0 +1,57 @@
...
+
+Ideally, our simulated samples should feature the same stripping cuts as the real data we want to work with.
+We can be sure of this if the same stripping version has been used when processing the simulated and real data.
+
+But often, our simulated sample will have a different version of the stripping applied to it.
+For example, what if our data sample uses Stripping 21, while our MC sample uses Stripping 20?
+
+In this case, we simply need to rerun our stripping line of choice from the correct stripping version.
+
+[This example](code/14-rerun-stripping/options.py) is an extended version of the minimal DaVinci DecayTreeTuple job that additionally runs the corresponding stripping line from Stripping 21.
```
betatim: Should we make "minimal DV DTT" link to that lesson?

index.md
```diff
@@ -28,7 +28,7 @@ Paragraph(s) of introductory material.
 8.  Switch to a new stripping (20 -> 21) for one line in MC
 9.  Split your jobs into subjobs
 10. (Stripping line efficiency)
-11. How to use a stripping line for a slightly different decay
+11. [How to use a stripping line for a slightly different decay](14-rerun-stripping.html)
```
betatim: Originally we thought that this line would be taking a candidate from an existing stripping line and replacing a mass hypothesis (eg. D->pipi gets changed to D->Kpi). So I would link this under line 28

ibab:
Yup, somehow picked the wrong line. I've fixed it and added a link to the DV example.

betatim:
LGTM.

betatim:
Looks good to me.

apuignav:
I have added PR #28  to remove the old Stripping bank so there's no need to put it in a weird place.

Issue #7, opened 2015-05-08
Minimal local DV job to make minimal DecayTreeTuple


betatim:
Given a local DST, run a minimal DV job over it to make a minimal DecayTreeTuple.

This will have to explain how to get stripping21 decisions (for the one chosen stripping line: http://lhcb-release-area.web.cern.ch/LHCb-release-area/DOC/stripping/config/stripping21/charmcompleteevent/strippingd2hhcompleteeventpromptdst2d2rsline.html) and use those as the file we download from the grid most certainly will have some "random" stripping

ibab:
Do you have something like this in mind?
https://twiki.cern.ch/twiki/bin/view/LHCb/StrippingFAQ#How_do_I_run_some_stripping_line

betatim:
Yes, that looks right.

The problem is that the simulated events you find in the book-keeping has a particular version of the stripping decisions already evaluated (but it keeps all events). So if we use `stripping21` for the data, then we should also use that for the simulated events.

alexpearce:
Is this for running over MC or data?

betatim:
A minimal DV job for data and for simulation.

In the data you can't redo the stripping, so the discussion of the stripping is for sim only

ibab:
Are there arguments against using Stripping20 for the tutorials?
Rerunning the stripping here makes the exercise a bit non-minimal :)

alexpearce:
I guess the argument against is that S21 is supposedly the ‘legacy stripping’, so if your data is available in S21 and you're starting a new analysis, that's what you should use. (I don't know how much of an official recommendation that is though.)

betatim:
I like the idea of using stripping21 because that is what everyone will be
using soon no? (Maybe this assumption is wrong) So by using it we make the
tutorial slightly more future proof. Maybe do a round of minimal DV
without, then have a second iteration where we say "and now with
stripping21"?

ibab:
Or maybe "this is how you access the stripping output" (using Stripping20) and then "oh, but we need to use Stripping21 -> rerun" in a second tutorial that follows this one?

Edit: Okay, realized that's identical to your suggestion :D

betatim:
@ibab can you write this bit? Pleeeaassee

alexpearce:
I'm already on it, but feel free to take over @ibab if you like.

ibab:
I can do the follow-up once I know how this part is structured / what the options file looks like.

betatim:
[...]

PR #19, opened 2015-05-19
Minimal DaVinci job run locally

alexpearce:
Addresses issue #7. Guides the student through creating an options file that will create an ntuple of D*+ candidates when run over a stripped DST.

### DIFF
--- a/09-minimal-dv-job.md
+++ b/09-minimal-dv-job.md
@@ -1,0 +1,155 @@
+---
+layout: page
+title: First steps in LHCb
+subtitle: Running a minimal DaVinci job locally
+minutes: 10
+---
+
+> ## Learning Objectives {.objectives}
+>
+> * Run a DaVinci job over a local DST
+> * Inspect the ntuple output
+> * Set up the job to run in Ganga
+
+With some stripped data located, it's useful to store the information on the 
+selected particles inside an ntuple.
+This allows for quick, local analysis with [ROOT](https://root.cern.ch/), 
+rather than always searching through a DST that contains lots of things we're 
+not interested in.
+
+As well as being the application that runs the stripping, 
+[DaVinci](http://lhcb-release-area.web.cern.ch/LHCb-release-area/DOC/davinci) 
+allows you to access events stored in DSTs and copy the information to ROOT 
+ntuples.
+You tell DaVinci what you want it to do through _options files_, written in 
+Python.
+There are lots of things you can do with DaVinci options files, as there's lots 
+of information available on the DST, but for now we'll just work on getting the 
+bare essentials up and running.
+
+Our main tool will be the `DecayTreeTuple` object, which we'll create inside a 
+file we will call `ntuple_options.py`:
+
+```python
+from Configurables import DecayTreeTuple
+from DecayTreeTuple.Configuration import *
+
+# Stream and stripping line we want to use
+stream = 'CharmCompleteEvent'
+line = 'D2hhCompleteEventPromptDst2D2RSLine'
+
+# Create an ntuple to capture D*+ decays from the StrippingLine line
+dtt = DecayTreeTuple('TupleDstToD0pi_D0ToKpi')
+dtt.Inputs = ['/Event/{0}/Phys/{1}/Particles'.format(stream, line)]
+dtt.Decay = '[D*(2010)+ -> (D0 -> K- pi+) pi+]CC'
+```
+
+This imports the `DecayTreeTuple` class, and then creates an object called 
+`dtt` representing our ntuple.
+Once DaVinci has run, the resulting ntuple will be saved in a folder within the 
+output ROOT file called `TupleDstToD0pi_D0ToKpi`.
+
+The `Inputs` attribute specifies where `DecayTreeTuple` should look for 
+particles, and here we want it to look at the output of the stripping line 
+we're interested in.
+
+As stripping lines can save many decays to a DST, the `Decay` attribute 
+specifies what decay we would like to have in our ntuple.
+If there are no particles at the `Input` location, or the `Decay` string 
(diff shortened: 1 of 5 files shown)


09-minimal-dv-job.md
```diff
@@ -0,0 +1,155 @@
...
+>
+> A microDST (or µDST) is a smaller version of a DST.
+> Some stripping lines go to µDSTs, and some go to DSTs.
+> There are two things that need changing it our options file in order to have 
+> in work when it is used with a stripping line that goes to a µDST:
+>
+> 1. The `DecayTreeTuple.Inputs` attribute should start at the word
+>    `Phys`; and
+> 2. The `RootInTES` attribute on the `DaVinci` object to be set to 
+>    `/Event/STREAM`
```
betatim: Make this $STREAM or some such to indicate it is a variable, not the literal word STREAM

alexpearce: :+1:

09-minimal-dv-job.md
```diff
@@ -0,0 +1,155 @@
...
+
+```python
+from Configurables import DaVinci
+
+# Configure DaVinci
+DaVinci().UserAlgorithms += [dtt]
+DaVinci().InputType = 'DST'
+DaVinci().TupleFile = 'DVntuple.root'
+DaVinci().PrintFreq = 1000
+DaVinci().DataType = '2011'
```
betatim: How did you find out that 2011 is the one to use for this MC sample? Worth mentioning as that seems like a FAQ

betatim:
If we add something about the 2011 thing I think this is ready to merge?

09-minimal-dv-job.md
alexpearce: If the MC DST is from 2012, this is wrong and should be `'2012'`.

alexpearce:
Just so I'm sure then, the DST these options will run over is an MC DST? If that's the case we will need to mention DDDB and CondDB tags.

09-minimal-dv-job.md
betatim: It is. The way to know is to look at the BK path. Can we add that sentence somewhere?

betatim:
The files I got for the book-keeping example are MC, full DST (I think).

alexpearce:
OK I'll add something then, though it is a little fiddly unfortunately.

betatim:
https://github.com/betatim/lhcb-first-analysis-steps/blob/gh-pages/data/MC_2012_27163003_Beam4000GeV2012MagDownNu2.5Pythia8_Sim08e_Digi13_Trig0x409f0045_Reco14a_Stripping20NoPrescalingFlagged_ALLSTREAMS.DST

We can take the info from this file no? Which also answers the question "How do I find my tags?"

alexpearce:
Yes, we can. I've had instances where that information wasn't present in the bookkeeping information though, so you have to try [alternative procedures](https://wiki.physik.uzh.ch/lhcb/davinci:dbtags).

One method is to use the Dirac API, which often produces the same result as in the bookkeeping `.py` file:

``` shell
$ SetupLHCbDirac
$ lhcb-proxy-init
$ dirac-bookkeeping-production-informations 35742
```

Where 35742 is the job production ID, the number after the `ALLSTREAMS.DST` part of the LFN with the leading zeros removed. For me, today, this produces

```
Production Info:
    Configuration Name: MC
    Configuration Version: 2012
    Event type: 27163003
-----------------------
 StepName: Merge14 for Sim08
    ApplicationName    : LHCb
    ApplicationVersion : v35r4
    OptionFiles        : $APPCONFIGOPTS/Merging/CopyDST.py
    DDB                : None
    CONDDB             : None
    ExtraPackages      :AppConfig.v3r164
-----------------------
Number of Steps   10
Total number of files: 20
         LOG:10
         ALLSTREAMS.DST:10
Number of events
File Type           Number of events    Event Type          EventInputStat
ALLSTREAMS.DST      140687              27163003            140687
Path:  /MC/2012/Beam4000GeV-2012-MagDown-Nu2.5-Pythia8/Sim08e/Digi13/Trig0x409f0045/Reco14a/Stripping20NoPrescalingFlagged
/MC/2012/Beam4000GeV-2012-MagDown-Nu2.5-Pythia8/Sim08e/Digi13/Trig0x409f0045/Reco14a/Stripping20NoPrescalingFlagged/27163003/ALLSTREAMS.DST
```

This shows the information for the merging production step, where all the individual DSTs are merged, and has `None` for both tags, so is unhelpful.

Unfortunately, the only foolproof way I know of getting these tags is to
1. Find out the production ID by inspecting the LFN;
2. Go to the [transformations monitor](https://lhcb-web-dirac.cern.ch/DIRAC/LHCb-Production/lhcb_user/Production/LHCbTransformationMonitor/display) and input the production ID in the “ProductionID” search field;
3. Click on the row that's returned as the search result and then “Show Request”, which will show the MC production request that created that production ID;
4. If the request has subrequests (as it does in this case), click the little “+” icons on the left of the row to expand it, and click the “+” icons again until you find the request that corresponds to the event type you're interested in (ID 20789 here);
5. Click on the row with the correct event type and then click “View”;
6. The first step shown, the step that ran Gauss, will show the tags.

For data, [the advice](https://twiki.cern.ch/twiki/bin/view/LHCb/RecommendedTags) is to just use the latest tags, so you don't need to go through all of this. For simulation it really is necessary: if you don't, you can end up with the wrong magnet polarity value in your ntuple, amongst other things.

alexpearce:
Pushed information on database tags in 6eb9a0d.

betatim:
I see another little lesson on how to find DB tags coming ...

If you fix the year in the options file, I am :+1: on merging.

alexpearce:
Done.

PR #20, opened 2015-05-20
Interactive DST exploring

betatim:


code/dst-explore.py
```diff
@@ -0,0 +1,133 @@
+# SetupProject DaVinci v36r6
```
alexpearce, line 1: I guess you know of `dst-explorer.py` script in Bender; is this script better?

betatim, line 1: I know of `dst-explorer.py` but it was too difficult/intransparent to use
with its lack of docs hence I started my own which just starts python,
opens the DST and then leaves you to do what you want to do. The Bender
script seems to do a lot of magic, takes/took "for ever" to start, had no
docs, ...

code/dst-explore.py
```diff
@@ -0,0 +1,133 @@
...
+        nodenames.append(node.identifier())
+
+    return nodenames
+
+def advance(decision='B02DKWSD2HHHBeauty2CharmLine'):
+    """Advance until stripping decision is true, returns
+    number of events by which we advanced"""
+    n = 0
+    while True:
+        appMgr.run(1)
```
apuignav, line 46:
``` python
if not evt["/Event/DAQ/RawEvent"]:
    print "Reached end of input files"
    break
```

betatim:
I haven't added the rendered HTML yet, will do as soon as we are happy (makes the diffs easier to read).

From my side it is ready to merge, otherwise it will become huge.

apuignav:
Merged. Let's work on this now.

betatim:
I added the rendered HTML directly to `gh-pages`.

alexpearce:
I was going to say: change “python” to “Python”.

PR #22, opened 2015-05-21
Update index.md

particleist:


betatim:
Closing, after switching to auto building HTML we can create a new PR for this. The target should be `master` then not `gh-pages`

PR #25, opened 2015-05-21
Add a few fixes to lesson 07-minimal-dv

ibab:
These are the fixes referenced in #7.


betatim:
paging soon-to-be doctor @alexpearce, the stream name difference is because MC <-> Data?

ibab:
Right, there's no /Event/CharmCompleteEvent in the MC DST

betatim:
LGTM, so we merge it then?

ibab:
Maybe Alex has a comment. We can always work on something else in the meantime :)

alexpearce:
Sorry for the delay. All looks good, thanks very much for the spotting the mistakes.

PR #27, opened 2015-05-21
Add a logo for the Starterkit.

kdungs:
Also minor modifications to the html templates.


betatim:
Can you update the "Source" link to point to this repo? What email should we put in the contact box?

kdungs:
Sure.
I would suggest yours or we create a mailing list that forwards everything to us.

betatim:
I suggest `lhcb-starterkit@cern.ch`, see also #29 about moving to an organisation.

betatim:
`lhcb-starterkit@cern.ch` is now a thing, so add it as contact

kdungs:
Changed and rebased. Also the manual handling of the compiled html is a hassle with git. What was the reason again we did it this way?

alexpearce:
It is annoying. I think because [GitHub pages can't build this site for us](https://github.com/swcarpentry/lesson-example#creating-your-lessons-repository) (point 8)?

kdungs:
Okay, so in principle we should be able to set up [Travis](https://travis-ci.org) to compile for us and push to gh-pages. Actual development should be done in another branch anyway.

ibab:
I've forgot to `make` or committed the wrong html lots of times already...

It shouldn't be too difficult to set something up that compiles all branches and pushes them to `gh-pages` into specific locations (With `master` being placed at `/`).
This way we could also quickly take a look at pull requests without checking out the branch locally.

kdungs:
Issue #30 created.

PR #28, opened 2015-05-21
Add EventNodeKiller for ReStripping

apuignav:
This way it's not necessary to put the Stripping reports elsewhere. Addresses concerns in #18 ,


alexpearce:
I would prefer not to do this as it complicates matters. Why does it matter if the stripping reports are not in the default location?

apuignav:
Well, if it's explained I don't see why it complicates matters. The idea
was to make sure all TupleTool worked out of the box, which was not clear
if the reports were not in the correct place.

alexpearce:
“Complicates” in that it adds more tools to run, that (looking at other peoples options files) will get copied blindly to newer files (though that is what we're trying to prevent, of course…!).

The only TupleTool that would need the stripping output is `TupleToolStripping`, right? Which has [`TupleToolStripping.StrippingReportsLocation`](http://lhcb-release-area.web.cern.ch/LHCb-release-area/DOC/davinci/latest_doxygen/d0/d4f/class_tuple_tool_stripping.html#a4255f29366e88fcbb005f354e36ce341) for setting a custom location. But, we're only capturing events which pass the stripping anyway aren't we? So we only care about that decision, which would be `1` for every event in the ntuple.

apuignav:
Well, my opinion is that this will cause more problems than solve them in case people copies the options blindly. Imagine a typical case in which a user is testing the efficiency of several Stripping lines, then `TupleToolStripping` will not work out of the box and this may create issues. In fact, it will work but with the old Stripping in the MC file, which is even worse.

I think this quite a well motivated inclusion of a tool that is going to be useful, plus it teaches a very interesting concept which is the Reports and the locations in the TES. If you want, I can expand a little bit  more the explanation to make that clear.

alexpearce:
OK, that makes sense. I would like some more motivation and explanation then. I think pretty much what you've said is good enough.

> The decisions of the stripping that ran during the central MC production are placed in a default location in the TES: `/Event/Strip/Phys/DecReports`. If we try to run our custom stripping, it will complain because the location it wants to write the decisions to will already exist. We run an instance of `EventNodeKiller` to remove the decisions from the MC production, so that our custom stripping can write there instead. This is nice, because most tools expect to read the stripping decisions from the default location, so we won't have to reconfigure anything.
> 
> Code…

apuignav:
That's a nice suggestion, I've updated the text accordingly.

kdungs:
> Make it so

PR #31, opened 2015-05-22
Remove unused references and adjust tests accordingly.

kdungs:
`make check` complained about missing pages. These were unused but marked as required in the test. I changed the requirements and removed the links from the index.


kdungs:
Ready for merge I think.

PR #32, opened 2015-05-22
Fix all errors raised by `make check`

kdungs:


ibab:
The `.prereq` thing is weird, no?
I can remember it showing up correctly on the page, so maybe we don't want to remove it completely?

kdungs:
Yeah, maybe the test needs to be fixed but then again it we agreed to just remove it.

PR #35, opened 2015-05-22
Plan for the last afternoon

betatim:
Added to the index, no actual material

### DIFF
--- a/32-install-party.md
+++ b/32-install-party.md
@@ -1,0 +1,89 @@
+---
+layout: page
+title: First Steps in LHCb
+subtitle: Install party
+minutes: 10
+---
+> ## Learning Objectives {.objectives}
+>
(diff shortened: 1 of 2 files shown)


ibab:
Looks good to me. Compiling `git` seems a bit overkill to me. Are there any big reasons against using the pre-installed `git` (1.7)?

betatim:
I think some of the instructions/guides/SO questions you find these days
won't work that well with an old version/there are some oddities (only one
I remember is default push strategy) between modern `git` and lxplus `git`.
Feels wrong to be on a different major version than the rest of the world.

I like campaigning for "only consensus is the most recent version of
everything" (too much time using gentoo).

ibab:
Okay, we can probably come back to this later on if/once we have more stuff to add here.

PR #39, opened 2015-05-23
[WIP] Minimal DaVinci on the grid

betatim:
Running the minimal DaVinci example on the grid.


betatim:
Premature merge, work continues in #40

Issue #9, opened 2015-05-08
Minimal DV job on the grid


betatim:
ganga is the interface to the grid, convert the minimal DV job from #7 to run on the grid, configure storing output on EOS, access the output file and TBrowser it to make mass plot

betatim:
This will pick up the files we selected and downloaded as `.py` file in #3 and use them as input. No need to explain `BKQuery`

PR #40, opened 2015-05-23
Minimal DV on the grid (continued)

betatim:
This is a continuation of #39


betatim:
Basically ready for review and then merging.

Should we add more about `ganga` here? New topic? Just linking?

betatim:
Fixes #9

ibab:
I usually run my job scripts from within `ganga` with `%ganga`.
Is that something people do?
I find this more convenient than closing/starting `ganga`, which takes quite a bit of time.

betatim:
Good point. I reworded it so that we just talk about the "inside ganga" option.

betatim:
Fixes #9

PR #41, opened 2015-05-23
Automating the creation of CERN cloud instances

betatim:
Still waiting to see if this actually works, creating instances is incredibly slow


betatim:
This seems to work though spinning up five instances at the same time doesn't. The fifth seems to fail with "can't find suitable host".

`create_instances.sh` could be nicer ...

I think this can be merged. For the workshop could y'all spin up four instances as well? Each instance only has 8GB of RAM and 4 CPUs, so not that powerful. Would need some coordinating in terms of hostnames.

betatim:
Filed a report [here](https://cern.service-now.com/service-portal/view-incident.do?n=INC0792661) about the instances not starting up.

ibab:
I think we should keep this PR open until all issues are resolved.
I would also like to add jupyterhub to the config.

betatim:
Trying jupyterhub is on my TODO list for today. Needs installing `python3`
(outside of anaconda) to run the hub, inside hopefully we can use `python2`
as a kernel with all the anaconda stuff.

ibab:
What worked well for me when I tried to come up with a way to set it up:
- Create Python3 environment in our anaconda installation:

```
$ conda create -n py3k python=3 anaconda
```
- Before setting up/running jupyterhub, activate it:

```
$ source activate py3k
```

ibab:
I have 5 machines running atm.
Everything seems to be working.
Are yours also running right now?

We should come up with a naming scheme so we can understand who owns which instances.
maybe lhcbstarterkitA{1..5}, lhcbstartkitB{1..5} ?

alexpearce:
For `iptables`: first get the default set of rules in to a file:

``` bash
$ iptables -L > iptables.up.rules
```

Edit this to include the Jupyter rule. The full file on my (not `lhcbstarterkit`) VM would look like

```
Chain INPUT (policy ACCEPT)
target     prot opt source               destination
ACCEPT     all  --  anywhere             anywhere            state RELATED,ESTABLISHED
ACCEPT     icmp --  anywhere             anywhere
ACCEPT     all  --  anywhere             anywhere
ACCEPT     tcp  --  anywhere             anywhere            state NEW tcp dpt:ssh
ACCEPT     udp  --  anywhere             anywhere            state NEW udp dpt:afs3-callback
ACCEPT     tcp  --  anywhere             anywhere            state NEW tcp dpt:4241
ACCEPT     tcp  --  anywhere             anywhere            state NEW tcp dpt:8000
REJECT     all  --  anywhere             anywhere            reject-with icmp-host-prohibited

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination
REJECT     all  --  anywhere             anywhere            reject-with icmp-host-prohibited

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination
```

Move it.

``` bash
$ sudo mv iptables.up.rules /etc/iptables.up.rules
```

Create a script to have `iptables` ‘source’ these rules when it starts up

``` bash
$ vim /etc/network/if-pre-up.d/iptables
```

fill it with

``` bash
#!/bin/sh
iptables-restore < /etc/iptables.up.rules
```

mark it as executable

``` bash
chmod +x /etc/network/if-pre-up.d/iptables
```

You can test that this works by flushing the current ruleset and reloading it from the file.

``` bash
$ iptables -F
$ iptables-restore < /etc/iptables.up.rules
```

apuignav:
Question: if we merge can I use this directly (ie, blindly) to bring instances up?

betatim:
You can already use it semi-blindly (read the included README ;)) by adding my fork as a remote and checking out the vm-setup branch (I can teach you git-fu if you need to).

I'd like to keep this open until we have a few hours of starterkit behind us and are sure nothing is missing. As the commits show we keep finding stuff that needs adding ...

Then there is the question of implementing the above suggestions from @alexpearce and @ibab. Though maybe I can play stupid and let them do it themselves after this has been merged :-/

betatim:
Igor can you circulate an email to "all of us" with a proposed numbering scheme and assignment? I like the `lhcbstarterkit[ABCD]N` idea.

I will restart all my instances except for `lhcbstarterkit8` later tonight.

The command to run once the machines are up (as root):

``` bash
$ PATH=/usr/local/bin:$PATH jupyterhub
```

Best done in a `screen`. Together with checking that the `iptables` rule in `configure-cm.sh` worked (pretty sure it does).

We will also have to add (by hand) the SWC instructors as users. I couldn't work out how to get `passwd` to change the local password. I edited `/etc/shadow` by hand to remove the `!!` and then `su -l` to the new user and set the password.

tools/cloud-machines/README.md
```diff
@@ -0,0 +1,9 @@
+Setting up VM on the CERN cloud
+===============================
+
+1. [Subscribe to the Cloud service](http://clouddocs.web.cern.ch/clouddocs/tutorial_using_a_browser/subscribe_to_the_cloud_service.html)
+2. Upload your public ssh key to the Cloud service [access and security](https://openstack.cern.ch/dashboard/project/access_and_security/)
+3. `source setup-lxplus.sh`
```
alexpearce, line 6: This file is called [`lxplus-setup.sh`](https://github.com/betatim/first-analysis-steps/blob/vm-setup/tools/cloud-machines/lxplus-setup.sh).

alexpearce:
I've started `lhcbstarterkit{20,21,22}`.

alexpearce:
The `kernelspec` stuff won't work because the Anaconda install is not known to the install script; `PATH` isn't modified.

alexpearce:
[ah nvm](https://github.com/lhcb/first-analysis-steps/pull/41/files#diff-222960ed2fdd232a01d4cd9b0b8f046aR38).

alexpearce:
It seems that only git 1.7.1 is available on the VMs. GitHub [recommends at least 1.7.10](https://help.github.com/articles/https-cloning-errors/), and older versions seem to have some problems pushing and pulling with HTTPS URLs.

alexpearce:
Are we going to use this for the [next Starterkit](https://indico.cern.ch/event/402899/)?

betatim:
From last year's experience I would say no. It seems people were successful in installing stuff on their laptops. We can take the [install instructions from SWC](https://twitwi.github.io/2015-06-02-cern-lhcb/) again?

alexpearce:
Yeah they look OK.

betatim:
I find the scripts in here useful despite not being quite finished. I will
keep the branch in my fork, should we also merge it here?

Otherwise close this. I created #139 for adding the instructions

alexpearce:
I would rather keep the repository clean, containing only what's needed for the course. So I'll close, but the code will stick around for reference.

PR #42, opened 2015-05-24
Prerequisites

betatim:
These are the steps to follow to make sure people are prepared for the workshop.

If I send them out by email, should we already link to them directly or copy and paste them in the email?

Is it a good idea to simply remove the "Please do not try and do this, but here are the instructions anyway" part at the end?

### DIFF
--- a/00-prerequisites.md
+++ b/00-prerequisites.md
@@ -6,22 +6,29 @@ minutes: 10
 ---
 > ## Learning Objectives {.objectives}
 >
-> * Check you are ready for the workshop
+> * You will be ready for the workshop!
 
 Please read and try the following steps **before** arriving. For
-some of the steps the solution is easy, but requires waiting for
-a day or so. So please try them before arriving and try to fix
-what ever does not work.
+some of the steps the solution requires waiting for a day or so.
+So please try them before arriving and try to fix what ever does
+not work.
+
+> Follow this guide before arriving, we will not have time to help
+> you with problems on these issues during the workshop. This means
+> you will end up watching instead of participating.
 
 This will be an interactive workshop, so you will need to bring
 a computer. There will be no machines for you to use in the room.
+
+> Follow all the steps using the computer you plan to bring, not
+> your dekstop or someone elses computer.
 
 If this is the first time you are bringing your laptop to CERN, you
 will have to [register it](https://network.cern.ch) before it can
 access the internet.
 
 Please bring an ethernet/network cable (even if you have wifi) and
-your power supply.
+your power supply, as well as a plug adaptor to swiss and european plugs.
 
 > In the following we assume you use Mac OS or Linux, unfortunately
 > we do not know how to do these things on Windows. If you do know,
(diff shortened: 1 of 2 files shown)


00-prerequisites.md
```diff
@@ -0,0 +1,58 @@
...
+Please bring an ethernet/network cable (even if you have wifi) and
+your power supply.
+
+> In the following we assume you use Mac OS or Linux, unfortunately
+> we do not know how to do these things on Windows. If you do know,
+> please contact us (lhcb-starterkit@cern.ch) so we can update these instructions.
+
+Try the following steps with the computer you will use at the workshop:
+
+1. In your browser, on the computer you are planning to bring to the
```
sredford: Suggest to remove "on the computer you are planning to bring to the workshop".

00-prerequisites.md
```diff
@@ -0,0 +1,58 @@
...
+   workshop try and access the [web based book keeping](https://lhcb-portal-dirac.cern.ch/DIRAC/).
+   If you need help with your grid certificate there is the
+   [Grid certificate FAQ](https://twiki.cern.ch/twiki/bin/view/LHCb/FAQ/Certificate)
+   and you can ask questions on `lhcb-distributed-analysis@cern.ch`.
+1. Connect to lxplus with `ssh -X lxplus.cern.ch`.
```
sredford: Should you mention which program to use? (Terminal on a Mac). Should you also mention that people might need to put their username first (eg sredford@lxplus.cern.ch)

00-prerequisites.md
```diff
@@ -0,0 +1,58 @@
...
+2. Once connected, check your grid certificate works by typing
+   `lhcb-proxy-init`. If you need help with your grid certificate
+   there is the [Grid certificate FAQ](https://twiki.cern.ch/twiki/bin/view/LHCb/FAQ/Certificate)
+   and you can ask questions on `lhcb-distributed-analysis@cern.ch`.
+3. Check that XForwarding works by running `xeyes` on lxplus. A set
```
sredford: Suggest to keep the language consistent: "by typing 'xeyes'."

00-prerequisites.md
```diff
@@ -0,0 +1,58 @@
...
+Try the following steps with the computer you will use at the workshop:
+
+1. In your browser, on the computer you are planning to bring to the
+   workshop try and access the [web based book keeping](https://lhcb-portal-dirac.cern.ch/DIRAC/).
+   If you need help with your grid certificate there is the
+   [Grid certificate FAQ](https://twiki.cern.ch/twiki/bin/view/LHCb/FAQ/Certificate)
+   and you can ask questions on `lhcb-distributed-analysis@cern.ch`.
+1. Connect to lxplus with `ssh -X lxplus.cern.ch`.
+2. Once connected, check your grid certificate works by typing
+   `lhcb-proxy-init`. If you need help with your grid certificate
```
sredford: What should happen?

00-prerequisites.md
```diff
@@ -0,0 +1,58 @@
...
+   there is the [Grid certificate FAQ](https://twiki.cern.ch/twiki/bin/view/LHCb/FAQ/Certificate)
+   and you can ask questions on `lhcb-distributed-analysis@cern.ch`.
+3. Check that XForwarding works by running `xeyes` on lxplus. A set
+   of eyes following your mouse should appear on your screen. Press
+   `Ctrl-C` to exit.
+4. Try and login to `lhcbstarterkit4.cern.ch` from an lxplus node. If
```
sredford: How do I do that? What is a node? (might be simpler to just say "From lxplus, try to logon..."

00-prerequisites.md
betatim: Added it to make clear that they need to try this on the computer they want to bring. It is no good if it works on their desktop but not the laptop they actually bring (you have to install stuff in the browser)

00-prerequisites.md
```diff
@@ -0,0 +1,58 @@
...
+   username (Find out your lxplus username by running `whoami` on lxplus) and
+   the error message you get.
+4. On `lhcbstarterkit4.cern.ch` run:
+   
+     1. `ipython --version` which should print `3.0.0`
+     2. `git --version` which should print `git version 2.4.1`
+   
+     If either of the above two commands do not work please email `lhcb-starterkit@cern.ch`
+
+During the workshop we will provide you with a machine to connect to from your
```
sredford: I would remove this 'during' sentence, it's unnecessary information which only serves to confuse.

00-prerequisites.md
```diff
@@ -0,0 +1,58 @@
...
+title: First Steps in LHCb
+subtitle: Pre-workshop Checklist
+minutes: 10
+---
+> ## Learning Objectives {.objectives}
+>
+> * Check you are ready for the workshop
+
+Please read and try the following steps **before** arriving. For
+some of the steps the solution is easy, but requires waiting for
```
sredford: This could be clearer. I would remove the "easy" part (what does that matter) and instead emphasise more that if people turn up without these things sorted they will be watching rather than participating.

00-prerequisites.md
```diff
@@ -0,0 +1,58 @@
...
+
+This will be an interactive workshop, so you will need to bring
+a computer. There will be no machines for you to use in the room.
+
+If this is the first time you are bringing your laptop to CERN, you
+will have to [register it](https://network.cern.ch) before it can
+access the internet.
+
+Please bring an ethernet/network cable (even if you have wifi) and
+your power supply.
```
sredford: What kind of plug/adapter does my power supply need?

00-prerequisites.md
sredford: In that case, I think it should be said in every point... Either you say it at the top, and it covers all points, or you say it in every point.

ibab:
This looks good. If we decide to support Windows, we should add instructions in a separate PR.

Issue #10, opened 2015-05-08
Add a simple TupleTool


betatim:
Add a tupletool to the minimal DV job from #9, re-run it, TBrowser to show new branches, how to find the code of the TupleTools, what tupletools exist?

ibab:
#45 is implementing this.

apuignav:
Closed by #45

PR #45, opened 2015-05-25
Add TupleTools to minimal example

apuignav:
I created the skeleton, now working on filling it.

### DIFF
--- a/12-add-tupletools.md
+++ b/12-add-tupletools.md
@@ -63,7 +63,8 @@ dtt.Decay = '[D*(2010)+ -> ^(D0 -> ^K- ^pi+) pi+]CC'
 This will run all the configured `TupleTools` on the marked particles, with the caveat that some tools are only run on certain types of particles (eg, tracking tools on particles that have an associated track).
 This configuration is not optimal, since there may be tools which we only want to run on the D's and some only on the children. Enter `Branches`, which allow us to specify which tools get applied to which particle in the decay (in addition to the `TupleTools` configured at the top level).
 
-Branches are configured by means of a `dict` that relates the variable name and the decay descriptor needed to select it. This also allows to give more descriptive names to the particles, which will be used as the prefix of the tuple leaves (of the form `PARTICLENAME_VARNAME`)
+Branches let you define custom namespaces in your ntuple by means of a `dict`:
+its keys define the name of each branch (and, as a consequence, the prefix of the corresponding leaves in the ntuple), while the corresponding values are decay descriptors that specify which particles you want to include in the branch.
 
 ```python
 dtt.addBranches({'Dstar' : '^[D*(2010)+ -> (D0 -> K- pi+) pi+]CC',
--- a/12-add-tupletools.md
+++ b/12-add-tupletools.md
@@ -14,7 +14,7 @@ minutes: 15
 
 Usually, the default information stored by `DecayTreeTuple` as shown in our [minimal DaVinci job](09-minimal-dv-job.html) is not enough for physics analysis. 
 Fortunately, most of the information we need can be added by adding C++ tools (known as `TupleTools`) to `dtt`;
-there is an extensive library of these, which will be briefly discussed later.
+there is an extensive library of these, some of which will be briefly discussed during the lesson.
 
 > ## Default DecayTreeTuple tools {.callout}
 > The default tools added in `DecayTreeTuple` are:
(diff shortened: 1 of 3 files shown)


apuignav:
At this point, I am thinking it would be a good idea to merge this lesson with "13. LoKi Hybrid tool", since it's a very small explanation we need to add and this one is not so long right now. Thoughts?

betatim:
What do you think about merging 13 and 14? They are both about LoKi.

apuignav:
I'm not even sure what we can explain in 14. Maybe it would be a good idea to have some "LoKi functors" lesson, if you want a little but more theoretical, and then this application on the DecayTreeTuple. That would be 13+14.

ibab:
Tim's #10 also suggested that we open TBrowser to take a look at the branches.
Maybe we should add this to the lesson, what do you think?

apuignav:
I am not sure how much we want here, so I added a very barebones explanation. I would also like them to explore a little bit on their own the clicky part.

12-add-tupletools.md
```diff
@@ -0,0 +1,117 @@
...
+means that the configured `TupleTools` will only run on the head of the decay chain, that is, the D*.
+In order to select the particles for which we want the information stored, we need to mark them with a `^` symbol in the decay descriptor.
+For example, if we want to fill the information of the D0 and its children, we would modify the `dtt` to look like this:
+
+```python
+dtt.Decay = '[D*(2010)+ -> ^(D0 -> ^K- ^pi+) pi+]CC'
+```
+
+This will run all the configured `TupleTools` on the marked particles, with the caveat that some tools are only run on certain types of particles (eg, tracking tools on particles that have an associated track).
+Clearly, this configuration is not optimal, since there may be tools which we only want to run on the D's and some only on the children. Enter `Branches`, which allow us to specify which tools get applied to which particle in the decay (in addition to the `TupleTools` configured at the top level).
```
betatim: All words ending in -ly should not be used :)

12-add-tupletools.md
```diff
@@ -0,0 +1,117 @@
...
+> - `DecayTreeTupleDalitz` for Dalitz analysis.
+> - `DecayTreeTupleJets` for obtaining information on jets.
+> - `DecayTreeTupleMC` gives us access to MC-level information.
+> - `DecayTreeTupleMuonCalib` for muon calibration tools.
+> - `DecayTreeTupleReco` for reconstruction-level information, such as `TupleToolTrackInfo`.
+> - `DecayTreeTupleTracking` for more detailed tools regarding tracking.
+> - `DecayTreeTupleTrigger` for accessing to the trigger information of the candidates.
+>
+> The `TupleTools` are placed in the `src` folder within each package and it's usually easy to get what they do just by looking at their name.
+> However, the best way to know what a tool does is check its documentation, either by opening its `.h` file or be searching for it in the latest `doxygen`.
```
betatim: turn doxygen into a link to the doxygen for DV v36r6

12-add-tupletools.md
```diff
@@ -0,0 +1,117 @@
...
+For example, if we want to fill the information of the D0 and its children, we would modify the `dtt` to look like this:
+
+```python
+dtt.Decay = '[D*(2010)+ -> ^(D0 -> ^K- ^pi+) pi+]CC'
+```
+
+This will run all the configured `TupleTools` on the marked particles, with the caveat that some tools are only run on certain types of particles (eg, tracking tools on particles that have an associated track).
+This configuration is not optimal, since there may be tools which we only want to run on the D's and some only on the children. Enter `Branches`, which allow us to specify which tools get applied to which particle in the decay (in addition to the `TupleTools` configured at the top level).
+
+Branches are configured by means of a `dict` that relates the variable name and the decay descriptor needed to select it. This also allows to give more descriptive names to the particles, which will be used as the prefix of the tuple leaves (of the form `PARTICLENAME_VARNAME`)
```
ibab: Maybe we should specifically spell out that the keys in the dict are the PARTICLENAMEs? (Might be worth calling them BRANCHNAMES, too)
This is probably very obvious, but it wouldn't hurt to write it down :)
As in 

> Branches let you define custom namespaces in your NTuple.
> They are configured by means of a `dict`.
> The keys of the `dict` let you define the name of each branch (Let's call these BRANCHNAME), while the corresponding values are decay descriptors that specify which particles you want to include in the branch.
> TupleTools can then be configured per branch and place their variables into `BRANCHNAME_VARIABLENAME`

apuignav:
True, I've changed the wording to make it clearer.

12-add-tupletools.md
```diff
@@ -0,0 +1,118 @@
...
+> However, the best way to know what a tool does is check its documentation, either by opening its `.h` file or be searching for it in the latest [`doxygen`](http://lhcb-release-area.web.cern.ch/LHCb-release-area/DOC/davinci/releases/latest/doxygen/index.html).
+> Most tools are very well documented and will also inform you of their configuration options.
+> As an example, to get the information on the `TupleToolTrackInfo` we used before we could either check its [source code](https://svnweb.cern.ch/trac/lhcb/browser/Analysis/trunk/Phys/DecayTreeTupleReco/src/TupleToolTrackInfo.h) or its [web documentation](http://lhcb-release-area.web.cern.ch/LHCb-release-area/DOC/analysis/releases/latest/doxygen/da/ddd/class_tuple_tool_track_info.html).
+> In case we need more information or need to know *exactly* what the code does, the `fill` method is the one we need to look at.
+
+The updated options `ntuple_options.py`, which can be found [here](./code/12-add_tupletools/ntuple_options.py) can be run in the same way is in the [minimal DaVinci job](09-minimal-dv-job.html) lesson.
```
ibab: is -> as

betatim:
LGTM, @ibab?

seneubert:
@ibab please consider using TreeViewer instead of TBrowser, it is a more useful interface in many respects

apuignav:
@sneubert maybe a separate lesson? They should already know ROOT...

seneubert:
Yes, we can put this into a separate lesson. I just don't want to promote the TBrowser.

apuignav:
Fair enough. I can just change this, but I don't know how to load the
TreeViewer from the command line (always do it from the TBrowser). Hints?

seneubert:
You can access the TreeViewer from TBrowser.
I usually do

``` c++
DecayTree->StartViewer()
```

from the root command line

ibab:
I'll be happy to merge this as soon as @apuignav says it's ready.

apuignav:
It's ready on my side.

Issue #14, opened 2015-05-08
Fun with LoKi functors


betatim:
Move the complicated stuff from #11 here. Then go even deeper?

How to look at the code of a LoKi functor, how to find more LoKi functors, how to figure out what they are doing, why do we need these "adaptor" functions, how to edit the code of a functor and use it.

apuignav:
Do we want to go deeper? I can write a small lesson on how to write a functor, but maybe it's too much and we should stick with #48.

betatim:
I will close this for now

Issue #11, opened 2015-05-08
LoKiHybrid TupleTool


betatim:
How to get access to LoKi functors to add new variables to the nTuple without writing c++

Here some simple and complex uses of LoKi functors:

``` python
jpsi_preamble = ["DZ = VFASPF(VZ) - BPV(VZ)",
              "TZ2 = DZ*3096.916 / PZ/299792458000.0",#in seconds
              "C = 299792458000.0",
              "EPZ = (PCOV2(3,3))**0.5",
              "ZERRV = (VFASPF(VCOV2(2,2)))**0.5",
              "ZERRPRIMARY = BPV(VCOV2(2,2))**0.5",
              "ZERR2 = ZERRPRIMARY*ZERRPRIMARY + ZERRV*ZERRV",
              ]
jpsi_vars = {"ETA": "ETA",
          "Y": "Y",
          "PHI": "PHI",
          "VPCHI2": "VFASPF(VPCHI2)",
          "DELTAZ": "DZ",
          # DZ * M / PZ / c with c in units of mm/s
          "TZ": "TZ2*(10**12)", #ps
          "TZERROR": "( (TZ2*TZ2*EPZ*EPZ)/(PZ*PZ) + (10**24*3096.916*3096.916*ZERR2)/(C*C*PZ*PZ) ) **0.5",
          #DZ*3096.916 / PZ/299792458000.0*(10**12)", #ps
          "minpt": "MINTREE('mu+' == ABSID, PT)",
          "minclonedist": "MINTREE(ISBASIC & HASTRACK, CLONEDIST)",
          "maxtrchi2dof": "MAXTREE(ISBASIC & HASTRACK, TRCHI2DOF)",
          }

tuple.addBranches({"JPsi": "^(%s)"%(decay_descriptor),
                 "muplus": mark(3, decay_descriptor),#"J/psi(1S) -> mu- ^mu+",
                 "muminus": mark(2, decay_descriptor),#"J/psi(1S) -> ^mu- mu+",
                 })
loki_JPsi = tuple.JPsi.addTupleTool("LoKi::Hybrid::TupleTool/LoKi_JPsi")
loki_JPsi.Variables = jpsi_vars
loki_JPsi.Preambulo = jpsi_preamble
```

apuignav:
Closed by #48

PR #48, opened 2015-05-26
LoKi functor lesson

apuignav:
Describe LoKi functors, types, etc, and how to use them. Trying to address #14 and #11.

### DIFF
--- a/13-loki-functors.md
+++ b/13-loki-functors.md
@@ -43,7 +43,7 @@ cand = cands[0]
 We can now try to get very simple properties of the B candidate, such as its transverse momentum or measured mass.
 
 ```python
-from LoKiPhys.functions import PT, MM
+from LoKiPhys.decorators import PT, MM
 print PT(cand)
 print MM(cand)
 ```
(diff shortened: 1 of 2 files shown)


ibab:
So far, the lesson describes what LoKi functors are in terms of their inputs/outputs and which categories of functors there are.
I think it would be nice to start with some motivation on why they're useful.
Like

> LoKi functors are designed to flexibly compute and compare properties of the current decay.
> For example, the functor 'PT' gives you the tranverse momentum of a particle.
> Internally, functors are implemented as C++ classes.
> ...

But I don't know that much about LoKi functors. What do you think?

apuignav:
Nice idea, it's really missing an introduction. Let's see how I implement
it.

betatim:
Some motivation would be good as well as some examples of doing stuff with them. From the easy ones like `PT` to trying to get info about a vertex associated to your particle where you have to use `BPV` and weird stuff like that.

What is the most ambitious/complicated use of LoKi functors you have ever seen? We should take that as a final puzzle and explain it at the very end.

apuignav:
Yes yes, this was just the starting point. Now I'll try to explain how to
use them using the `explore-dst` session.

On Tue, May 26, 2015 at 1:35 PM, Tim Head notifications@github.com wrote:

> Some motivation would be good as well as some examples of doing stuff with
> them. From the easy ones like PT to trying to get info about a vertex
> associated to your particle where you have to use BPV and weird stuff
> like that.
> 
> What is the most ambitious/complicated use of LoKi functors you have ever
> seen? We should take that as a final puzzle and explain it at the very end.
> 
> —
> https://github.com/lhcb/first-analysis-steps/pull/48#issuecomment-105494841
> .

## 

Dr. Albert Puig Navarro
Laboratoire de Physique des Hautes Energies
Ecole Polytechnique Fédérale de Lausanne (EPFL)
BSP 614.4 (Cubotron UNIL) CH-1015 Lausanne
EPFL Phone: 021 6939808
CERN Phone: 72518

apuignav:
Right now I have the basic LoKi functors. The problem is that showing them in an interactive GaudiPython (the easiest) gives some nasty errors + we cannot use the BPV functors. Ideas?

apuignav:
So, should we go deeper? Which are your favorite functors to include?

ibab:
This is awesome!
Can you add a link to the example code and an invitation to try it out at the very end?

apuignav:
Should be OK now.

apuignav:
I think `SUMTREE` and `ABSID` were important. Added and ready to merge.

ibab:
Looks good!
I'm merging this.

PR #50, opened 2015-05-26
Lesson on splitting Ganga jobs

ibab:


ibab:
What does everyone think about adding the following infos to the lesson?
- How to choose the number of files per job
- How to deal with jobs that are stuck

betatim:
Yes and yes.

Personal opinion on "files per job", as few as possible :-/

Worth introducing people to the `ganga` queues (`queues.add(j.submit)` and you get your `ganga` shell back instantly while the (sub)jobs submit in the background)

apuignav:
Well, if you're running on a full Stripping, the least amount of files per
job may be a problem (too long submission time). The balance is important.

betatim:
I think we should add a callout box mentioning `queues.add(j.submit)` (with caveat that you have to keep `ganga` running while it works through the queue). It is a useful thing to know about and comes up most often in context of many subjobs.

Another motivation for splitting your job is that if half your files are at CERN and half at SARA you won't be able to process them all unless you allow your job to be split up.

ibab:
I've implemented the suggestions.
Should be good to merge from my side.

betatim:
Thanks for this! :cake:
Merged

Issue #46, opened 2015-05-26
CombineParticles/Selection framework lesson


apuignav:
Should we have a lesson on CombineParticles/FilterDesktop, etc?

seneubert:
I will include a very basic bit on selections for the changeing of mass hypothesis

betatim:
You need to do selection framework stuff to change mass hypos?

apuignav:
I don't think so, I would do a separate lesson, when I finish LoKi functors
I can do it.

betatim:
It would make sense to have a small explanation on what it is and how stripping lines come into existence. Writing a stripping line isn't a beginner task though, so some part of me is thinking "beyond the scope of this tutorial". Maybe it fits well in the "general outline about stripping" part? Was actually wondering what to put there.

seneubert:
Yes you need a selection to do change the mass hypo. I'll open a new issue to discuss the solution

alexpearce:
Closed by #71.

Issue #23, opened 2015-05-21
Automatically run our examples


betatim:
We should find some way to make our documentation "executable". Unsure yet how to do that as it will require ganga.

PR #53, opened 2015-05-26
Switching mass hypothis

seneubert:
I added the solution of a DV script that changes a mass hypothesis. Please have a look at code/18-switch-mass-hypo/ntuple_switchHypo.py

I also added a template for the corresponding lessen.

### DIFF
--- a/code/18-switch-mass-hypo/ntuple_switchHypo.py
+++ b/code/18-switch-mass-hypo/ntuple_switchHypo.py
@@ -1,0 +1,69 @@
+from GaudiConf import IOHelper
+from Configurables import DaVinci, DecayTreeTuple
+from DecayTreeTuple.Configuration import *
+
+# Stream and stripping line we want to use
+stream = 'AllStreams'
+line = 'D2hhCompleteEventPromptDst2D2RSLine'
+tesLoc = '/Event/{0}/Phys/{1}/Particles'.format(stream, line)
+
+# prepare a Filter algorithm to creat a new Selection of particles
+from Configurables import FilterDesktop
+alg = FilterDesktop('SelFilter')
+alg.Code = 'ALL' # we could filter out some data here. For now just select ALL candidates
+
+# get the selection(s) created by the stripping
+from PhysSelPython.Wrappers import Selection
+from PhysSelPython.Wrappers import SelectionSequence
+from PhysSelPython.Wrappers import DataOnDemand
+
+strippingSels = [DataOnDemand(Location=tesLoc)] 
+
+# instantiat a Selection that uses the algorithm defined above
+# and gets the stripping selection as input
+sel = Selection('Dst2D0pi_D0Kpi_Sel', Algorithm=alg, RequiredSelections=strippingSels)
+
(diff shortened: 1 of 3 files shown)


code/18-switch-mass-hypo/ntuple_switchHypo.py
```diff
@@ -0,0 +1,69 @@
...
+alg.Code = 'ALL' # we could filter out some data here. For now just select ALL candidates
+
+# get the selection(s) created by the stripping
+from PhysSelPython.Wrappers import Selection
+from PhysSelPython.Wrappers import SelectionSequence
+from PhysSelPython.Wrappers import DataOnDemand
+
+strippingSels = [DataOnDemand(Location=tesLoc)] 
+
+# instantiat a Selection that uses the algorithm defined above
```
betatim: instantiat -> instantiate

betatim:
From what I can tell the code looks alright. Does it run? (We need #23)

For the lesson we should come up with a good "excuse" why we are doing this, so that people can relate to it. Otherwise it might appears as some advanced gaudi selection kongfu with no purpose.

We should explain the why of the various steps that need to be done. Why a `FilterDesktop` that doesn't seem to do anything, why all these sequences of selection wrappers, what is a `GaudiSequencer`? Or asked differently: how do I know that I need to do these steps? (Otherwise people will just blindly copy the recipe)

apuignav:
I think we should have a lesson before introducing all the concepts. This
will help focusing this session on the mass hypothesis substitution.

betatim:
So #46 is going to be a real thing then. Will you write it?

apuignav:
As soon as I finish LoKi functors, which may take a while if I do it
properly. I will open a PR as soon as I start, so if someone has time to
start before, please feel free to do it.

seneubert:
This code runs. 
You are right, though that for our example Dst -> D0pi it is not very well motivated. I believe for the D0 reco D02pipi is even done in the stripping line already :(
But do we have Dst->D0(pipi)pi MC available?

I have an example where I do D+ --> Lambda_c+ but there is probably no MC for that.

If you know another good use case that would be simple to implement let me know. On the other hand I think there is nothing wrong with putting forward academic examples and let people figure out by themselves when it will become useful.

code/18-switch-mass-hypo/ntuple_switchHypo.py
```diff
@@ -0,0 +1,69 @@
...
+from DecayTreeTuple.Configuration import *
+
+# Stream and stripping line we want to use
+stream = 'AllStreams'
+line = 'D2hhCompleteEventPromptDst2D2RSLine'
+tesLoc = '/Event/{0}/Phys/{1}/Particles'.format(stream, line)
+
+# prepare a Filter algorithm to creat a new Selection of particles
+from Configurables import FilterDesktop
+alg = FilterDesktop('SelFilter')
```
seneubert: I am not convinced that we need this inital Filter here.
Let me try to simplify this.

seneubert:
The selection wrappers I copied.
I have no Idea why Gaudi works this way. Should I spend time to research this for the lesson?
Can I merge now?

PR #56, opened 2015-05-27
Add detailed Windows SSH instructions

lbel:
These have been tested on Windows XP, but contain nothing that shouldn't work on newer versions of Windows.

### DIFF
--- a/00-prerequisites.md
+++ b/00-prerequisites.md
@@ -30,9 +30,9 @@ access the internet. Both wifi and ethernet connection.
 Please bring an ethernet/network cable (even if you have wifi) and
 your power supply, as well as a plug adaptor to swiss and european plugs.
 
-> In the following we assume you use Mac OS or Linux, unfortunately
-> we do not know how to do these things on Windows. If you do know,
-> please contact us (lhcb-starterkit@cern.ch) so we can update these instructions.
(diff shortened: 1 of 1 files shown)


00-prerequisites.md
```diff
@@ -88,3 +88,24 @@ Proxies uploaded:
       If either of the above two commands do not work please email `lhcb-starterkit@cern.ch`
 
 If you can successfully execute all of the above steps you are ready to go for the workshop.
+
+> ## Windows-specific Instructions {.callout}
+>
+> On Windows, some steps are required before you can connect via SSH.
+>
+> Setup steps (you only have to perform this once):
+> 1. Download the [Xming installer](http://sourceforge.net/projects/xming/files/latest/download).
```
betatim, line 97: add a space here so the listing works properly

betatim:
Thanks a lot for this!!

The instructions look good to me, but then I haven't used windows in about a decade :-/

I would suggest we merge it unless someone is against it.

/cc @kdungs @alexpearce

kdungs:
I can try this at home on my Windows but from here it looks good.

betatim:
Next time can we wait for the formatting to be fixed? How do we repair this now?

PR #57, opened 2015-05-27
Add some content to general Gaudi/DaVinci introduction.

kdungs:
Refers to issue #4. Needs some more input from someone who actually knows something about Gaudi.


ibab:
Would you mind if I added some content to the branch?
You would need to add me as a collaborator to your fork (This would have been a reason to create branches in the main repo :/ )

kdungs:
Yes please! Adding you now.

ibab:
I'm not sure how all of this will work during the workshop.
I think it's important to explain these concepts up front, otherwise the rest of the workshop will be very opaque ("Why am I editing this Python file and what does it do, anyway?").

On the other hand, this is all pretty abstract and boring :/

apuignav:
I think it's quite a nice introduction and it's necessary to give a context, otherwise it's all tricks.

kdungs:
The changes you made make perfect sense and you got rid of the subheadings, perfect.

ibab:
I think this can be merged now.
We should improve the two lessons later on, but it's probably better to open a pull request for each one.

01-davinci.md
```diff
@@ -0,0 +1,82 @@
...
+> ## Learning Objectives {.objectives}
+> * Which basic problems is the LHCb software framework designed to solve?
+> * Learn the key concepts needed to work with the LHCb software
+>
+
+A HEP experiment like LHCb needs to solve many computational challenges, some of which are
+
+ - How do we collect data as it is recorded by the detector?
+ - How do we filter and process the recorded data efficiently?
+ - How do we manage all the complex tasks required to work with particle events?
```
seneubert: What is a `particle event` ? 
I'd rather not use that name but somewhere explain how an event is defined and how it relates to collisions and bunch crossings.

01-davinci.md
```diff
@@ -0,0 +1,82 @@
...
+ - How do we organize all the data of a single particle event in a flexible way?
+ - How do we configure our software flexibly without having to recompile it?
+
+At LHCb, we base our software on the [Gaudi](https://proj-gaudi.web.cern.ch/proj-gaudi/) framework, which was specifically designed with the above questions in mind.
+In order to understand how the LHCb software works, you should know about the following basic Gaudi concepts:
+
+ - Because the individual collisions are almost completely independent of each other, it makes sense to process them one by one (as a *stream*), without holding them all in memory at once.
```
seneubert: This is not quite true. We assume that bunch crossings (a.k.a events) are independent. But collisions can actually be correlated (accidentally) if you assign a particle to the wrong PV.

01-davinci.md
```diff
@@ -0,0 +1,82 @@
...
+   Gaudi provides a global *EventLoop*, which runs over individual events, and allows you to process them one by one
+ - A single event contains lots of different data objects (*Particles*, *Vertices*, *Tracks*, *Hits*, ...).
+   In Gaudi, these are organized in the *Transient Event Store* (TES).
+   You can think of it as a per-event file system with locations like `/Event/Rec/Track/Best` or `/Event/Phys/MyParticles`.
+   These usually contain containers of C++ objects.
+   When running over the event stream, Gaudi allows you to `get` and `put` to/from these locations, like a key-value store.
```
seneubert: to/from --> from/to

kdungs:
:+1: on merge

01-davinci.md
ibab: I would rather avoid going into a definition of particle events here.
Maybe _collision data_ would be more appropriate?
This line is supposed to be the motivation for having `Tools` in Gaudi.

01-davinci.md
ibab: I meant to say bunch crossings here. Will change this.

betatim:
Can we move the discussion of "developing LHCb software" to a later part of the lesson? Right now we don't need it for any of the lessons. `lb-run` is enough. It feels a bit like introducing several abstract things, that are not used in any of the topics later and because it is at the start of the day, it is using up some of the most valuable time.

Compare "DecayTreeTuple is a Gaudi algorithm which is executed in DaVinci which is based on the Gaudi framework and it uses DecayTreeTupleTools to write individual branches" with "I want to make an nTuple, for this you use the DecayTreeTuple algorithm which you can run in DaVinci." I prefer the latter when trying to persuade people that this is useful. The other just bamboozles with big words (if you aren't a software kind of guy).

ibab:
Yes! I just put one lesson after the other when splitting them.
Maybe we should put the dev lesson right before the how to ask questions lesson?

One danger of starting off in full  "I want to make an ntuple" mode is that people will not get the big picture of what the software does or how it works, considering it is quite complex and opaque.
Things like "the code in option files is not actually executed, it just sets stuff in the C++" need to be explained at some point.

Even the SWC people have detailed introduction chapters: http://swcarpentry.github.io/shell-novice/00-intro.html

ibab:
Okay, this needs a new review.

16-lhcb-dev.md
```diff
@@ -0,0 +1,101 @@
...
+
+inside the directory. This will (similar to `lb-run`) give you a new bash session with the right environment variables.
+
+Your new development environment won't be very useful without any software to modify and build.
+So why not check out out one of the existing LHCb packages, which are stored in the LHCb SVN repository?
+
+Let's assume you have already used `lb-dev` to set up your development environment and you are currently inside it.
+In order to obtain the source code of the package you want to work on, use `getpack`.
+This is an LHCb-aware wrapper around SVN.
+For example, if you want to write a custom stripping selection, write
```
betatim: ", write:" -> ", execute the following in the `DaVinciTest` directory:

16-lhcb-dev.md
```diff
@@ -0,0 +1,101 @@
...
+
+```bash
+getpack Phys/StrippingSelection
```
betatim: Should we tell them about `HEAD` and a specific version? One of the most confusing things is that if you `getpack` a specific version, wanting to use that as your starting point, then make edits, then decide you want to commit them you are stuck. You first have to do some magic to go to the `HEAD` version, re-apply your changes, commit them. I bet a lot of people don't bother doing that (for sure I often can't be bothered for small changes). What is a workflow that reduces this level of tediousness? `getpack HEAD`, `svn update $rev_of_v34r5`, do my stuff, jump to the head and `svn commit`? Unsure.

16-lhcb-dev.md
```diff
@@ -0,0 +1,101 @@
...
+> Alternatively, you can perform an anonymous checkout:
+> ```
+> getpack -p anonymous Phys/StrippingSelection
+> ```
+
+Under the hood, `getpack` will `svn checkout` (≈ `git clone`) the corresponding SVN repository.
+If you have made changes that are supposed to be integrated into the official LHCb repositories, you can use `svn commit`.
+Be advised that you should always communicate with the package maintainers before committing changes!
+
+If you just want to take a look at a source file, without checking it out, you can comfortably access the repository through two different web UIs:
```
betatim: add "(which one to use? It is just a matter of taste, pick the one that looks nicest)"

betatim:
I am happier now with both the lessons and their placement. However I had a lot of opinions so we should get someone else to say something too.

01-davinci.md
```diff
@@ -0,0 +1,119 @@
...
+
+**Transient Event Store**
+A single event contains lots of different data objects (*Particles*, *Vertices*, *Tracks*, *Hits*, ...).
+In Gaudi, these are organized in the *Transient Event Store* (TES).
+You can think of it as a per-event file system with locations like `/Event/Rec/Track/Best` or `/Event/Phys/MyParticles`.
+When running over the event stream, Gaudi allows you to `get` and `put` from/to these locations.
+
+**Algorithms**
+An *Algorithm* is a C++ class that can be inserted into the EventLoop.
+These allow you to perform a certain function for each bunch crossing (like filtering events, reconstructing particles, ...).
```
apuignav: I don't like bunch-crossing here. I think it should be very clear that the Gaudi processing is event-by-event and that Algorithms run _once_ per event.

01-davinci.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+During this run, DaVinci didn't do anything: We didn't specify any algorithms to run or any data to run over.
+Usually, you will write an option file (e.g. `options.py`) and specify it as an argument to `gaudirun.py`:
+
+```bash
+lb-run DaVinci v36r6 gaudirun.py options.py
+```
+
+An `option.py` is just a regular Python script that specifies how to set things up in the software.
```
apuignav, line 95: Maybe add you can specify several options files and they will be used in sequence to configure Gaudi. This is very useful when you build resuable components which just have small differences (data/MC, for example) that can be extracted in separate options files.

16-lhcb-dev.md
apuignav: I would just leave it at HEAD and explain that otherwise they can getpack a specific version but they won't be able to commit to it. I think the code here should be changed to getpack head, though.

betatim:
@ibab still working on the other line notes?

ibab:
Which ones still need to be addressed in your opinion?

My battery's empty, I'll get back to working on lessons as soon as I can
plug it in somewhere :)

betatim:
@apuignav your bunch crossing comment, has that been addressed?

The other comment has been addressed but my lazy brain just assumed it hadn't because it wasn't marked as "comment on outdated diff".

ibab:
I think the bunch crossing one is the main one remaining. Are we really
sure that Gaudi processes the proton collisions individually? I always had
the impression that we have access to the entire bunch crossing on each
Algorithm execution.

apuignav:
Yes, we access event by event, so bunch crossing per bunch crossing. When I
read the text this was confusing with the previous points in the sense that
event and bunch crossing are used interchangeably, even in the same
sentence.

betatim:
I have no strong opinion on this. For me bunch crossing == event ...

apuignav:
Agreed, but let's use only one or express clearly their equivalence. Instead of complaining, I paste here my suggestion for the final text (with a couple extra changes):

**Event Loop**
Because the individual bunch crossings (events) are almost completely independent of each other, it makes sense to process them one by one, without holding them all in memory at once.
Gaudi provides a global _EventLoop_, which allows you to process events one by one.

**Transient Event Store**
A single event contains lots of different data objects (_Particles_, _Vertices_, _Tracks_, _Hits_, ...).
In Gaudi, these are organized in the _Transient Event Store_ (TES).
You can think of it as a per-event file system with locations like `/Event/Rec/Track/Best` or `/Event/Phys/MyParticles`.
When running over the event stream, Gaudi allows you to `get` and `put` from/to these locations.
The contents of the TES are emptied at the end of the processing of each event.

**Algorithms**
An _Algorithm_ is a C++ class that can be inserted into the EventLoop.
These allow you to perform a certain function for each event (like filtering according to trigger decision, reconstructing particles, ...).

**Tools**
Often, algorithms will want to make use of some common function (vertex fitting, calculating distances, associating a primary vertex, ...).
These are implemented as _Tools_, which are shared between Algorithms.

betatim:
@ibab or @kdungs can you rebase this and deal with Albert's comments?

After that I think this is good to be merged.

ibab:
Okay, should be good to merge now.

PR #63, opened 2015-05-28
Use a different TES location to detect file end

### LINKED ISSUE #51: Detection of end of file doesn't work in 05-interactive-dst.md
While this usually works, the lines

``` python

if not evt["/Event/DAQ/RawEvent"]:
    print "Reached end of input files"
    break

```

in `05-interactive-dst.md` actually don't work in this file because `/Event/DAQ/RawEvent` doesn't exist. I don't know why, I have been using for years.

betatim:
Fixes #51 

Added a caveat to explain that this is not a fool proof method and might fail if you are looking at a file that was not processed by `Brunel`.

### DIFF
--- a/05-interactive-dst.md
+++ b/05-interactive-dst.md
@@ -135,7 +135,7 @@ def advance(decision):
     while True:
         appMgr.run(1)
 
-        if not evt["/Event/DAQ/RawEvent"]:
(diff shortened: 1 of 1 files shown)


ibab:
Okay, I've tested it and it works with our DST :)

PR #64, opened 2015-05-28
Update references to SetupProject where possible

### LINKED ISSUE #62: SetpProject vs lb-run
We should make a decision and make sure we are consistent everywhere.

If we advertise `lb-run` we need to show off the `lb-run Project v1r2 $SHELL` trick. (Bonus points for making it modify the prompt to show what project you are in automagically ...)

ibab:
This fixes #62


betatim:
:chocolate_bar:

PR #65, opened 2015-05-28
Kerberos in screen

### LINKED ISSUE #36: Kerberos tokens in screen session
Documentation is incomplete.

betatim:
Added description of how to keep your kerberos tokens alive. This is based on an empirical study to find this method. There is also some `kerbpg.sh` script or some such, though I never understood what it does/why it should work.

Fixes #36 reported by @kdungs so he should have a look if this satisfies him

### DIFF
--- a/02-screen.md
+++ b/02-screen.md
@@ -118,3 +118,19 @@ $ screen
 # now inside the screen session
 $ kinit
 ```
+
+> ## Finding lost screens {.callout}
+>
+> Once you start a `screen` session you need to remember which
+> `lxplus` node it is running on. If you forget to note that down
+> you can use the following little snippet to find any `screen`
+> sessions running on `lxplus` nodes:
+>
+> ```bash
+> for i in $(seq -f "%04g" 1 500); do
+>  ssh -o ConnectTimeout=10 -o PreferredAuthentications=gssapi-with-mic,gssapi -o GSSAPIAuthentication=yes -o StrictHostKeyChecking=no -o LogLevel=quiet lxplus$i.cern.ch "(screen -list | head -1 | grep -q 'There is a screen on') && hostname && screen -list"
+> done
+> ```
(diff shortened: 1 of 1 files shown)


betatim:
Actually don't like that I don't know if this recipe actually works or just appears to work.

kdungs:
What I usually do is

``` bash
pagsh.krb -c 'kinit && tmux new-session
```

so for you it would be

``` bash
pagsh.krb -c 'kinit && screen'
```

also you will have to renew your token every once in a while (24h I think).

betatim:
This is the strategy I use all the time, and it appears to work.

Do you understand what `pagsh.krb` actually does?

Renew every 24h is unfortunately a thing.

apuignav:
I have a small snippet to find a `screen` when you've lost it in lxplus.

``` bash
for i in $(seq -f "%04g" 1 500); do
     ssh -o StrictHostKeyChecking=no -o LogLevel=quiet lxplus$i.cern.ch  -o ConnectTimeout=10 "hostname && ps faxu | grep `whoami` | grep SCREEN | grep -v grep" | grep -v "No route" | grep -v "Permanently" | grep -B1 SCREEN
done
```

Maybe it's interesting.

kdungs:
Do you use certificate based login for lxplus? Maybe I am overlooking something but right now I don't understand how this would not ask me for my password 500 times.

apuignav:
Argh, actually it doesn't, but I don't remember why. Just run it from lxplus :-)

kdungs:
Ah right that makes sense. Once you have a valid kerberos token that takes care of it.

betatim:
I added a slightly modified version of @apuignav snippet. I find it a little more "obvious" as to what it does. Added the preferred auth options to make sure it uses kerberos tokens, no matter what your `.ssh/config` says.

apuignav:
Nice improvement!

PR #66, opened 2015-05-28
Shuffled around LoKi functors lesson

apuignav:
I think everything is more logical now.


betatim:
Can you make a three line summary of the text changes? I was expecting to see just `index.md` being modified and some file renames but the text has also changed. Should I try and follow them?

betatim:
Another point, should we have a new PR for renaming the two topics that are `05-..md` right now are can we do that in this PR as well? If yes, we should rename the PR to something like "The great re shuffle of lesson numbering" or so.

apuignav:
Basically no text changes except what's necessary so nothing sounds weird.
I would just ask if the text flows better with the sessions split this way,
so the TupleTool one is longer and the LoKi one shorter.

The idea is that LoKi is more standalone and we can build other lessons
upon it.

Albert

apuignav:
I would wait until the end. Let's merge this one and once we're happy with
the lessons and their order we rename them.

However, we should think about dropping the numbers in front of the lessons
because every time we move or change something it's quite a mess.

12-add-tupletools.md
```diff
@@ -104,15 +105,63 @@ The usage of `Branches` is very important (and strongly encouraged) to keep the
...
+> root [1] TBrowser *b = new TBrowser()
+> root [2]
+> ```
+> 
+> Try to locate the branches we have added, which are placed in the `TupleDstToD0pi_D0ToKpi/DecayTree`, and plot some distributions by double-clicking the leaves.
+
+Picking up with the [LoKi functors lesson](06-loki-functors.html), let's store some specific bits of information discussed there in our ntuple.
+To add LoKi-based leaves to the tree, we need to use the `LoKi::Hybrid::TupleTool`, which is configured with 3 arguments:
+
+  - Its *name*, specified in the `addTupleTool` call after a `/`.  This is very useful (and recommended) if we want to have different `LoKi::Hybrid::TupleTool` for each of our branches. For instance, we may want to add some information on the D*, the D0 and the soft $\pi$:
```
betatim: .. add some information on ... -> ... add different information for ...

betatim:
After that minor comment of mine: LGTM

Once it is fixed I will merge it.

betatim:
This is a cool lesson, I learnt something! (And I think it will help people realise that LoKi isn't black magic)

apuignav:
This should be OK now.

betatim:
Merci! :pizza: and :wine_glass:

alexpearce:
This is very cool :smile: thanks!

Issue #13, opened 2015-05-08
Efficiency of a stripping line


betatim:
How do I calculate the stripping efficiency using signal MC? -> TupleToolStripping

alexpearce:
Use of `TupleToolStripping` comes with the caveats that:
1. The stripping reports in the MC are for the same stripping version as in your data
2. I don't _think_ prescales are applied to stripping lines in MC, so you need to take that efficiency in to account
3. Generally, you need to do something else if you have PID cuts in your stripping line

alexpearce:
Although on 1 you can rerun any stripping lines and specify the location of the selection reports, so you could give `TupleToolStripping` this location and it should give you the correct decisions.

betatim:
What is an alternative way of doing this to avoid `TupleToolStripping`?

We should gear everything towards the fact that we re-run the stripping for
our line of choice unless the bookkeeping suddenly has stripping21
"flagged" simulation samples.

betatim:
Regarding (3), what do you do? Maybe we should just use that as the default. Following the idea of presenting just one (obvious) way of doing things.

alexpearce:
Unfortunately it somewhat depends: either you care about the reconstruction efficiency, in between the generator level cuts and the stripping, or you don't.

In the latter case you just run the stripping on your MC and count how much (truth-matched) signal you have in your ntuple. The ratio (N_stripped/N_accepted) gives you the combined reconstruction and stripping efficiency. (N_accepted is the number passing the generator level cut defined in the decay file.)

In the former case you also need access to the candidates before the stripping, which you can do with `CombineParticles` and the `mcMatch` functor. Then the reconstruction efficiency is (N_reconstructed/N_accepted) and the stripping efficiency is (N_stripped/N_reconstructed). Often the product of these two is equal to (N_stripped/N_accepted), but I'm not wise enough to say under what circumstances.

For 3, you can copy the stripping module that defines your stripping line and alter it to remove PID cuts. If the PID cuts are completely configured by the line's configuration dictionary then you can just change that without needing the copy Python files around. (So only if you use the `NoPIDs` containers.)

alexpearce:
I guess though it's more useful to show them the techniques. Deciding what specifics things to use in their analyses is a problem they need to tackle themselves.

betatim:
You will write this bit then? :smile:

I think we should pick a way of doing it that isn't too simplistic (maybe also not the most complicated one) and show that off with a note (as usual): other methods are available

[...]

PR #76, opened 2015-06-01
Removing the stripping line efficiency topic

betatim:
For now let's remove it from the index. We can come back to this later on and add it back in.


betatim:
Implements #13

PR #78, opened 2015-06-01
Add an introduction to the homepage

alexpearce:
There wasn't one, so this is something. Couldn't think of much more to say…

Got rid of the callout because it's just repeating the prerequisites lesson. Would it be worth adding a sentence saying “make sure you've followed this lesson”, or is that obvious enough given that it's the first lesson?

### DIFF
--- a/index.md
+++ b/index.md
@@ -11,6 +11,15 @@ the knowledge gained from the previous ones.
 If you have any problems or questions, you can [send an email to 
 `lhcb-starterkit@cern.ch`](mailto:lhcb-starterkit@cern.ch).
 
+> ## Prerequisites {.prereq}
+>
+> Before starting, you should be familiar with using a shell, like `bash`, and 
+> with programming in Python.
+>
+> The [Software Carpentry 
+> workshop](http://twitwi.github.io/2015-06-02-cern-lhcb/) gives a great 
+> introduction to these, as well as to many other useful computing tools.
+
 ## Topics
 
 1.  [Pre-workshop checklist](00-prerequisites.html) 
--- a/index.md
+++ b/index.md
@@ -2,12 +2,14 @@
 layout: lesson
 title: First Steps in LHCb
 ---
-Paragraph(s) of introductory material.
 
-> ## Prerequisites {.prereq}
->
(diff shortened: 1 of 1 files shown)


apuignav:
I'm afraid it wants 1 and exactly 1 callout with prereq...

alexpearce:
Looks like `make check` [requires there to be at least one `prereq` box on the homepage](https://github.com/lhcb/first-analysis-steps/blob/master/tools/check.py#L459)…

Shall I just put something like “follow the prerequisite lesson”, or would people prefer some more detail? (Like “ssh to lxlus” etc.)

betatim:
I think the idea of the prereq box is to mention also other technical things. We could add a link to the SWC lessons on bash and python. We kind of assume you know your way around the shell and can read python

alexpearce:
Good idea. I'll add a link to our SWC page.

alexpearce:
Done.

apuignav:
Looks good to me.

On Mon, Jun 1, 2015 at 11:09 AM, Alex Pearce notifications@github.com
wrote:

> Done.
> 
> —
> https://github.com/lhcb/first-analysis-steps/pull/78#issuecomment-107376642
> .

## 

Dr. Albert Puig Navarro
Laboratoire de Physique des Hautes Energies
Ecole Polytechnique Fédérale de Lausanne (EPFL)
BSP 614.4 (Cubotron UNIL) CH-1015 Lausanne
EPFL Phone: 021 6939808
CERN Phone: 72518

betatim:
Language lawyering: should we call the whole `first-analysis-steps` thing a lesson and each of the pages a topic? Or each of the pages are a lesson each and then we call the whole thing ...(a workshop?)?

betatim:
Otherwise I am happy. Thanks for tidying.

alexpearce:
I would ultimately leave that to an executive decision :wink: For me, each individual step is a lesson, but maybe we should use the SWC lingo.

betatim:
Do we have any text that needs fixing if we decide one way or the other? If not my executive decision tool(tm) says: Heads

betatim:
Merging this for now, we can adjust the language and make it more consistent later.

PR #70, opened 2015-05-29
Add checking of code directory to `make check`

kdungs:
As a first step just check all the files in the code directory. Next step will be checking of inline python snippets.

PR #79, opened 2015-06-01
Pep8 ify

alexpearce:
Related to #70, fixes (nearly) all of the warnings that [`pep8`](https://pypi.python.org/pypi/pep8) emits on everything in `code/`, and mirrors those changes in the `.md` lessons. I've tried to make the coding style consistent.

There's a whole bunch o' warnings from `data/` and `tools/` that I haven't tried to deal with.  The remaining stuff in `code/` is

```
code/dst-explore.py:108:80: E501 line too long (102 > 79 characters)
code/dst-explore.py:110:80: E501 line too long (86 > 79 characters)
code/11-davinci-grid/first-job.py:4:80: E501 line too long (187 > 79 characters)
code/12-split-jobs/first-job.py:4:80: E501 line too long (187 > 79 characters)
```

I'm not sure how to break these lines up. We could append [`# noqa`](https://github.com/jcrocholl/pep8/pull/136) to the lines, but then those comments would need to be explained, else they'll be copied.

### DIFF
--- a/05-files-from-grid.md
+++ b/05-files-from-grid.md
@@ -49,7 +49,8 @@ Since those tuples tend to be quite large, you might want to use your AFS work d
 If you want to obtain all the files, you can copy and paste the list of file names from the file you got from the bookkeeping and paste them into the following python script for convenience.
 
 ```python
-FILES = []  # Your list of file names here
+# Your list of file names here
+FILES = []
 
 if __name__ == '__main__':
     from subprocess import call
@@ -61,9 +62,9 @@ if __name__ == '__main__':
 
     files = FILES[:n_files]
     for f in files:
-        print('Getting file %s.' % f)
-        call('dirac-dms-get-file %s' % f, shell=True)
-    print('Done getting %d files.' % n_files)
+        print('Getting file {0}.'.format(f))
+        call('dirac-dms-get-file {0}'.format(f), shell=True)
+    print('Done getting {0} files.'.format(n_files))
 ```
 
 Save it as `getEvents.py` and use it via `python getEvents.py [n]`. If you specify `n`, the script will only get the first n files from the grid.
--- a/05-interactive-dst.md
+++ b/05-interactive-dst.md
@@ -31,16 +31,20 @@ import sys
 
 import GaudiPython as GP
 from GaudiConf import IOHelper
-from Configurables import LHCbApp, ApplicationMgr, DataOnDemandSvc
-from Configurables import DecodeRawEvent
-from Configurables import CondDB
-from Configurables import DaVinci
+from Configurables import (
+  LHCbApp,
+  ApplicationMgr,
+  DataOnDemandSvc,
+  DecodeRawEvent,
+  CondDB,
+  DaVinci
+)
 
 
 appConf = ApplicationMgr()
 
 dv = DaVinci()
-dv.DataType = "2012"
+dv.DataType = '2012'
 
 dre = DecodeRawEvent()
 dre.DataOnDemand = True
@@ -99,7 +103,7 @@ def nodes(evt, node=None):
         for l in evt.leaves(node):
             # skip a location that takes forever to load
             # XXX How to detect these automatically??
-            if "Swum" in l.identifier():
+            if 'Swum' in l.identifier():
                 continue
 
             temp = evt[l.identifier()]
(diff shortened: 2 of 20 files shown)


betatim:
We could break up the long strings and rely on string continuation (or what ever the jargon is for "blah" "blah" on two lines being combined into "blahblah"). I think that is easier to explain than `# noqa`

alexpearce:
That would mean this:

``` python
j.inputdata = j.application.readInputData('data/MC_2012_27163003_Beam4000GeV2012MagDownNu2.5Pythia8_Sim08e_Digi13_Trig0x409f0045_Reco14a_Stripping20NoPrescalingFlagged_ALLSTREAMS.DST.py')
```

in to something like this:

``` python
j.inputdata = j.application.readInputData((
    'data/'
    'MC_2012_27163003_Beam4000GeV2012MagDownNu2.5'
    'Pythia8_Sim08e_Digi13'
    '_Trig0x409f0045_Reco14a_Stripping20NoPrescalingFlagged'
    '_ALLSTREAMS.DST.py'
))
```

split somewhat arbitrarily.

The decay descriptors are slightly harder to split, personally. How about from this:

``` python
# works
# decay_desc = '[[B0]cc -> (^D- => {^K- ^K+ ^pi-, ^K- ^pi+ ^pi-,^pi+ ^pi- ^pi-, ^K- ^K- ^pi+}) ^K-]cc'
# doesn't work
decay_desc = '[[B0]cc -> (^D- => {^K- ^K+ ^pi-, ^K- ^pi+ ^pi-,^pi+ ^pi- ^pi-}) ^K-]cc'
```

to this?

``` python
# works
decay_desc = (
    '[[B0]cc -> '
    '(^D- => {^K- ^K+ ^pi-, ^K- ^pi+ ^pi-,^pi+ ^pi- ^pi-, ^K- ^K- ^pi+})'
    '^K-]cc'
)
# doesn't work
decay_desc = (
    '[[B0]cc -> (^D- => {^K- ^K+ ^pi-, ^K- ^pi+ ^pi-,^pi+ ^pi- ^pi-}) ^K-]cc'
)
```

betatim:
Both BK path and decay descriptor look fine.

Don't think there is an ideal solution. Maybe we add a comment to each to explain what is happening for those not deep into python parser grammar?

alexpearce:
Maybe it's taught in one of the SWC lessons?

alexpearce:
`pep8 code` now emits no warnings.

I would like to get this merged before the lessons, as having a consistent coding style (done to a standard) is important.

seneubert:
Where are these DecayDescriptors coming from? They look suspiciously like old style ones, no? Also charges are wrong.

Shouldn't it be

``` python
decay_desc = (
    '[ B0 -> '
    '^(D- -> {^K- ^K+ ^pi-, ^K- ^pi+ ^pi-,^pi+ ^pi- ^pi-, ^K- ^K- ^pi+})'
    '^K+]CC'
)
```

Also, please note that this notation is looking super strange for anybody not very familiar with python. Please make a comment in the code if you use it.

Would like to merge because I think the pep8 stuff is good. But with these issues remaining we can't stop here. Ideas?

alexpearce:
I just formatted the descriptors that were there. If they are wrong we should change them, I agree. I would have we could get rid of them completely though, seeing as they are unrelated to the D*+ chain we'll be testing with.

The notation is first [explicitly explained in the Building Decays](https://github.com/lhcb/first-analysis-steps/pull/79/files#diff-2f3939c27f5f02116dd8f2e6d8b384c8R82) lesson.

You're right not to merge if there are outstanding issues.

betatim:
Let;s get rid of them. They are only in `dst-explorer.py` because I dumped a script I had there as a starting point. That part is never used in an y of the examples, we maybe can even get rid of `dst-explorer.py` all together?

If you add a commit removing the lines (I'd keep the file for the moment as I don't know off the top of my head if it is used or not) then we can merge this.

Sorry for not realising earlier that we were chasing a red herring, too busy trying to think how to solve the problem instead of big picture.

(Yes some of them are old style, at the time what I was doing was trying to understand how the decay finder tools work with old and new)

alexpearce:
Agree to keep for now. I've removed the lines and tidied the decay descriptor finding stuff.

betatim:
Thank you for making travis' brain even smarter! :slot_machine:

alexpearce:
we just need #70 to add the check to the tests.

betatim:
Think my comment about `pep8 .` in #70 is wrong, should be `pep8 code`

//Paging @kdungs

PR #81, opened 2015-06-01
Use dst-explorer for 05-interactive-dst

pieterdavid:
I had a look at a few of the tutorials (they look very nice; I will try to find a bit of time to contribute), and with the "interactive dst" one I had the feeling that this is a lot more convenient with the bender tools, so I went through it and replaced the gaudipython snippets with dst-explorer commands, and removed a few things (explanation of the loadDesktop warning, catching the end of the file) that are not necessary in this case... apologies if I entirely missed the point.

### DIFF
--- a/05-interactive-dst.md
+++ b/05-interactive-dst.md
@@ -23,99 +23,38 @@ is not working as expected.
 The file we [downloaded from the grid](05-files-from-grid.html)
 contains simulated data, with stripping and trigger decisions
 and so on. Here we assumed the file you downloaded is called `00035742_00000002_1.allstreams.dst`.
-To take a look at the contents of the TES we need a simple
-options file:
-
-~~~ {.python}
-import sys
-
-import GaudiPython as GP
-from GaudiConf import IOHelper
-from Configurables import LHCbApp, ApplicationMgr, DataOnDemandSvc
-from Configurables import DecodeRawEvent
-from Configurables import CondDB
-from Configurables import DaVinci
-
-
-appConf = ApplicationMgr()
-
-dv = DaVinci()
-dv.DataType = "2012"
-
-dre = DecodeRawEvent()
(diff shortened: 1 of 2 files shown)


saschastahl:
I am in favour of not using Bender. The old tutorial is also easily applicable to running a simple gaudipython script in Brunel, Moore etc when you want to look at stuff created inside the job. 
Of course if you want to simply look at a DST lying somewhere this is faster. Maybe keep both?

apuignav:
IMO, I think we should not veer away from `GaudiPython` in the basic tutorial. The point here is not to show how to do things, for which `Bender` has super nice shortcuts, but for the students to understand how some things work. I believe part of this is loading your own file, looking for an event that passes the stripping decision you want (which helps you realize not all the events have what you want in the TES) and to directly get things from the TES.
Additionally, GaudiPython works in other contexts, such as `Moore` or `Brunel`, which is something you cannot do in `Bender`.

betatim:
Thanks for the contribution!

As the others explained we chose to use simple python instead of using the tools in `Bender` on purpose. The point is to give people a feel for what is "inside" a DST, and that they can interact with it without the need of any complicated code. An interactive `python` session is enough.

There are many powerful tools in LHCb world that provide a high level of abstraction. This is a good thing. The problem with most of them is that they are not very well documented, they have (necessarily) lots of layers to them and can be fragile. This makes them more of an expert tool than something for beginners. For beginners you need either "high level, never need to look inside with super docs" or "show everything!".

There are several ideas and topics already for `second-analysis-steps` a second tutorial for people who have done this one. We should start a repo for that and start collecting there. I think this would be a better fit there. A second way of interacting with DSTs, after learning that in principle you could do things yourself.

(Following the logic of my favourite "poem":
...
Explicit is better than implicit.
Simple is better than complex.
...
)

PR #82, opened 2015-06-02
Add instructions for changing default shell

alexpearce:
The default LHCb shell is `tcsh`. This PR adds instructions to the prerequisites lesson for changing the default from `tcsh` to `bash`.


apuignav:
Yes please! Merged.

PR #83, opened 2015-06-02
Work on section "Developing LHcb software"

saschastahl:
I added some content to this section, like release notes, tag colloector and nightly tests. 
It might need some more editing and explanation.

### DIFF
--- a/16-lhcb-dev.md
+++ b/16-lhcb-dev.md
@@ -19,14 +19,6 @@ If you want to make changes to a software package, you will need to set up a dev
 lb-dev --name DaVinciDev DaVinci v36r6
 ```
 
-<!-- This doesn't seem to work atm
-Alternatively, if you want to use the most recent nightly builds:
-
-```bash
-lb-dev --name DaVinciDev --nightly lhcb-head DaVinci
-```
--->
-
 The output should look similar to this:
 
 ```
@@ -95,9 +87,6 @@ It works similar to `lb-run`, without the need to specify a package and version:
 > getpack -p anonymous Phys/StrippingSelection
 > ```
 
-If you have made changes that are supposed to be integrated into the official LHCb repositories, you can use `svn commit`.
-Be advised that you should always communicate with the package maintainers before committing changes!
-
 If you just want to take a look at a source file, without checking it out, you can comfortably access the repository through two different web UIs.
 
  * [Trac](https://svnweb.cern.ch/trac/lhcb/)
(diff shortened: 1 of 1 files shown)


16-lhcb-dev.md
```diff
@@ -106,5 +95,51 @@ If you just want to take a look at a source file, without checking it out, you c
...
+and try to build the given project.
+You can checkout this version by doing:
+```bash
+lb-dev DaVinci HEAD --nightly lhcb-head 
+```
+A more detailed description of the command is found here:
+
+ * [SoftwareEnvTools](https://twiki.cern.ch/twiki/bin/view/LHCb/SoftwareEnvTools)
+
+Sometimes mistakes happen and the commit is broken. Therefore the nightly tests are performed.
```
alexpearce: “commit is broken” → “build is broken”.

saschastahl:
I wanted to say that the committed code is wrong. That the build is broken the next logical step. I added a sentence to distinguish between a non compiling build and software which does not what it is supposed to do.

apuignav:
Shall I merge this? I don't have any complaints...

betatim:
Ok with me. Though I think this lesson isn't interactive enough. We should not say "go to the tag collector and add stuff" and then not have them do it. Similarly with the nightlies, we need to do something with them, show them how to actually run the tests that they need to check the output of with their changes.

(I understand it is difficult to setup things so we can have people put things in the tag collector, etc)

betatim:
Otherwise this becomes more a of a documentation than a lesson.

alexpearce:
We could go over the tag collector in the (forthcoming) SVN lesson #88?

betatim:
Yes, but then we should have this before the developing lesson.

Overall I am now approximately -0 on adding more lessons. I think we should
see how what we have goes, then adopt that and really start a
`second-analysis-steps` repo for more advanced things. There is an infinite
amount of stuff we could add, but we want to keep `first-analysis-steps`
really a set of lessons for beginners. If we keep making it bigger and more
advanced (even at the end) that makes it a worse set of lessons. Much
better and maintainable if we create a second lesson with this stuff.

seneubert:
I agree with Tim on not adding more advanced material at the moment.
If we still have time to work on the lessons we should rather improve the existing ones, in particular to make them interactive and to focus them on the most important and pragmatic issues, while generalizations are pushed into callout boxes.

saschastahl:
My plan was to commit stuff during the lesson and go through all steps. So it should be interactive.

betatim:
Ok, do you know what repository we can use for having 40 people making
commits to?

T

alexpearce:
Use the SVN docs. Each user area can be written by everyone.

Issue #84, opened 2015-06-02
Bug in the selection script


seneubert:
In https://github.com/lhcb/first-analysis-steps/blob/master/code/06-building-decays/build_decays.py#L14
the right property of CombineParticles is not the `Decay` but `DecayDescriptor`.
Have we tested this code snippet?

seneubert:
Also, in https://github.com/lhcb/first-analysis-steps/blob/master/code/06-building-decays/build_decays.py#L21 you cannot pass Pions and Kaons. These objects are no Selections. I suspect you have to use the DataOnDemand service to get them. Testing this out now.

apuignav:
Yes, I'm fixing the text. PR incoming...

On Tue, Jun 2, 2015 at 4:23 PM, Sebastian Neubert notifications@github.com
wrote:

> Also, in
> https://github.com/lhcb/first-analysis-steps/blob/master/code/06-building-decays/build_decays.py#L21
> you cannot pass Pions and Kaons. These objects are no Selections. I suspect
> you have to use the DataOnDemand service to get them. Testing this out now.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/lhcb/first-analysis-steps/issues/84#issuecomment-107971772
> .

## 

Dr. Albert Puig Navarro
Laboratoire de Physique des Hautes Energies
Ecole Polytechnique Fédérale de Lausanne (EPFL)
BSP 614.4 (Cubotron UNIL) CH-1015 Lausanne
EPFL Phone: 021 6939808
CERN Phone: 72518

seneubert:
Good fix.

PR #86, opened 2015-06-02
Bugfixing in selections lesson

apuignav:
Addressed bug #84.


seneubert:
Looks good to me.

apuignav:
So if Travis accept it, I would say let's merge. We need to start finishing up.

PR #109, opened 2015-06-04
Add substitutions to DecayTreeFitter lesson

seneubert:
script done and working.
ToDo: update lesson

This also fixes a bug in code/18-switch-mass-hypo/

### DIFF
--- a/22-decay-tree-fitter.md
+++ b/22-decay-tree-fitter.md
@@ -84,9 +84,36 @@ tv__tree->Draw("Dstar_ConsD_D0_M[0]>>h(100,1800,1900)","","", 128, 0);
 
 As expected, the D0 candidates are forced onto their PDG mass value.
 
-The solution to this exercise `ntuple_DTF1.py`, is [available 
-here](./code/22-decay-tree-fitter/ntuple_DTF1.py).
 
 > ## Explore {.challenge}
 > * Look at the `status` variable to check if the fits converged.
 > * Look at the chi2 distribution of the fit 
+
+`DecayTreeFitter` can be told to change some of the hypotheses in the decay tree. This is very useful if you want to slightly change, which decays you want to look at. As an example let's say we want to examine the cabbibo-suppressed decay of the D0 into pi- pi+ instead of K- pi+. For this we add a second fitter, giving it a new name `ConsDpipi`:
+```python
+dtt.Dstar.addTupleTool(TupleToolDecayTreeFitter('ConsDpipi'))
+dtt.Dstar.ConsDpipi.constrainToOriginVertex = True
+dtt.Dstar.ConsDpipi.Verbose = True
+dtt.Dstar.ConsDpipi.daughtersToConstrain = ['D0']
+```
+We now can tell the fitter to substitute the kaon in the D0 decay by a pion.
+```python
+dtt.Dstar.ConsDpipi.Substitutions = {
+    'Charm -> (D0 -> ^K- pi+) Meson': 'pi-',
+    'Charm -> (D~0 -> ^K+ pi-) Meson': 'pi+'
+}
+```
+In the dictionary that is passed to the `Substitutions` property of the fitter, the keys are decay descriptors, where the respective particle to be substituted is marked with a `^`. The values are the repective new particle hypotheses. The substitution will only work if you start from a decay descriptor that actually matches your candidates. However, you are allowed to generalise parts of the decay. Here we replaced `D*(2010)` with the more general `Charm` and the bachelor `pi-` is just represented by a `Meson`.
+ 
+Note that the substitution mechanism does not understand the `CC` symbol. Both charge states have to be specified explicitely.
+
+Running the ntuple script again with these additions gives you fit results for the re-interpreted decay. 
+
+> ## Challenge {.challenge}
+> * Compare the outcome of the two fits with the different mass hypothesis
+> * Compare the fit quality between the correct and the the wrong hypothesis 
+
+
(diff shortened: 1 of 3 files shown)


seneubert:
This is ready. Please have a look.

22-decay-tree-fitter.md
```diff
@@ -84,9 +84,36 @@ tv__tree->Draw("Dstar_ConsD_D0_M[0]>>h(100,1800,1900)","","", 128, 0);
...
-The solution to this exercise `ntuple_DTF1.py`, is [available 
-here](./code/22-decay-tree-fitter/ntuple_DTF1.py).
 
 > ## Explore {.challenge}
 > * Look at the `status` variable to check if the fits converged.
-> * Look at the chi2 distribution of the fit 
\ No newline at end of file
+> * Look at the chi2 distribution of the fit 
+
+`DecayTreeFitter` can be told to change some of the hypotheses in the decay tree. This is very useful if you want to slightly change, which decays you want to look at. As an example let's say we want to examine the cabbibo-suppressed decay of the D0 into pi- pi+ instead of K- pi+. For this we add a second fitter, giving it a new name `ConsDpipi`:
```
alexpearce: “change,” → “change” (remove comma).

“cabbibo” → “Cabibbo”.

22-decay-tree-fitter.md
```diff
@@ -84,9 +84,36 @@ tv__tree->Draw("Dstar_ConsD_D0_M[0]>>h(100,1800,1900)","","", 128, 0);
...
+```python
+dtt.Dstar.addTupleTool(TupleToolDecayTreeFitter('ConsDpipi'))
```
alexpearce: With `addTupleTool` you shouldn't pass an _instance_ of the tool, instead you should pass a string, e.g.

``` python
dtt.Dstar.addTupleTool('TupleToolDecayTreeFitter/ConsDpipi')
```

This is quite nice because the students learnt “instantiate a tool with a string, named with the stuff after the slash” in the interactive DST lesson, when using the tool service.

So, we should remove this line as well:

``` python
from Configurables import TupleToolDecayTreeFitter,TupleToolDecay
```

`addBranches` adds a `TupleToolDecay` instance to each branch so that you don't need to, and `addTupleTool` will do the instantiation of TupleTools for you.

22-decay-tree-fitter.md
```diff
@@ -11,9 +11,9 @@ minutes: 10
 > * Apply a mass constraint
 > * Inspect the refitted decay tree
 
-Once you have made a hypothesis on the chain of decays that have lead to your final state, you then can incorporate the additional knowledge that comes with this hypothesis to get a new best estimate for the particle parameters - in particular their momenta. The additional knowledge is represented as constaints, which your decay tree has to fulfill.
+Once you have made a hypothesis on the chain of decays that have lead to your final state, you then can incorporate the additional knowledge that comes with this hypothesis to get a new best estimate for the particle parameters - in particular their momenta. The additional knowledge is represented as constraints, which your decay tree has to fulfill.
```
alexpearce: “that have lead to your final state” → “that lead to your final state”.

“parameters - in” → “parameters — in” (sorry :smile: the former is a hyphen, the latter is an em-dash, which is more grammatically correct here)

apuignav: +1 for the em dash :-)

alexpearce:
This lesson in very nice :smiley: I like it a lot!

I'm not sure about this paragraph:

> For example, for the decay
> 
> ``` python
> '[D*(2010)+ -> (D0 -> K- pi+) pi+]CC'
> ```
> 
> you make the assumption that the (K- pi+) combine to form a D0 with a specific invariant mass. This results in a so called mass-constraint. In addition the kaon and the pion are supposed to originate from exactly the same point in space. If you are looking for prompt D\* you can require them to come from the primary vertex. Boundary conditions like those are called vertex-constraints.

It might be more clearer to say “you _can_ make the assumption…” (emphasis just for this comment), because of course by default this assumption has not already been made. Also, “If you are looking for prompt D_…”, might it be better to say “if you know that the D_ candidates originate from the primary vertex, you can constrain them to do so” or something. My hesitation with the sentence as it is now is that if you're _looking_ for the prompt D_, constraining to the PV  won't mean your D_ actually _did_ come from the PV.

And just one point I'm not sure about: it's said that “the kaon and the pion are supposed to originate from exactly the same point in space”, but there's no explicit (configurable) constraint made for this (unlike `constrainToOriginVertex`, of example). Is this just something DTF will do for you automatically, always assuming the hypothesis that the decay products originated from the same point in space?

I might reword to something like this:

> For example, for the decay
> 
> ``` python
> '[D*(2010)+ -> (D0 -> K- pi+) pi+]CC'
> ```
> 
> you can make the assumption that the `K- pi+` pair always has the D0 invariant mass. This results in a so-called mass constraint, and can improve the mass resolution of higher decay products, like the D\* in our case.
> 
> In addition, the kaon and the pion should to originate from exactly the same point in space. If you know that your data contains only prompt D*, you could require them to come from the primary vertex. Boundary conditions like these are called vertex constraints, and can improve the momentum resolution for the particles because some uncertainties on the vertex positions are reduced.

alexpearce:
Slightly off-topic: we can just teach the substitution stuff in here rather than lesson 16 (“Replace a mass hypothesis”) right?

seneubert:
Thanks for reviewing!
Implemented the changes.
Script is running. 
Merge at your convenience.

betatim:
Merging. B2 sends you a :tropical_drink:

PR #111, opened 2015-06-05
[WIP] Remove unmerged lessons and lessons we do not need

betatim:
Removing lessons that are not yet merged or that
we decided to get rid of/merged into other lessons.

Do we want to also reshuffle the index in this PR? I'd prefer to do it in its own PR


alexpearce:
This PR is outdated as these lessons are [no longer in the index in `HEAD`](https://github.com/lhcb/first-analysis-steps/blob/4da24014f9b5d19cdb630d4e40b434c60994010a/index.md), so I'll close.

Issue #108, opened 2015-06-04
Actual lesson schedule


betatim:
The schedule of how we actually did the lessons on day one:
1. Introduction to the LHCb Software
2. Using screen to keep things running
3. Find the dataset in the Bookkeeping
4. Download a file from the grid
5. Interactively exploring a DST
6. Fun with LoKi Functors (stopped before `CHILD` in B2)
7. Introduction to the LHCb data flow (B2 spent a lot of time on Stripping webpage)
8. Running a minimal DaVinci job locally
9. Storing large files on EOS
10. Minimal DaVinci on the grid

betatim:
We should update `index.md` to reflect this.

The plan for Friday is:
1. Show the success of the grid jobs
2. TupleTools and branches
3. Decay tree fitter with mass substitution added
4. Developing LHCb software
5. Asking good questions and helping people
6. Contributing to these lessons

// cc @kdungs is this what we wrote down on the paper?

seneubert:
We did "Asking good questions and helping people" while waiting for jobs  from 3 to finish.

PR #127, opened 2015-06-19
Remove lessons we did not use and renumber lessons

betatim:
With the creation of `second-analysis-steps` we don't need to keep the lessons we did not keep around anymore. They now all have their own repository.

Renumbered the lessons so that the filenames match the actual order.


betatim:
Used #108 as basis for what we covered

betatim:
_bump_

/cc @alexpearce @ibab @apuignav

ibab:
This looks good.

PR #131, opened 2015-07-01
Fixed the doxygen link to the LoKi Cuts namespace.

SherabWangchuk:
The link to the LoKi Cuts namespace doxygen was somehow sometimes broken (I remember Albert trying to access it during the LoKi functors lesson). So I replaced the links in lesson 06 and 12.


ibab:
Thanks!

PR #134, opened 2015-07-08
Add more details on DTF, which variables are affected when applying s…

FedericaLionetto:
…ome constraints and how to add the DTF information for the daughters of the daughters


apuignav:
Mmm, I would not add the reference to `DTF_FUN`, since it performs the refitting evey time and therefore it's slow if (ab)used. I think we want to keep this simple, so I would stick with the `TupleTool` only.

I let others comment (@sneubert), but I seem to remember we decided against it the first time for this reason.

apuignav:
Also, thanks for the PR!

FedericaLionetto:
Yes, I agree it's more complicated and that's actually why I didn't describe it explicitly.
But still, I think putting a link there shouldn't hurt too much.
I've been struggling with how to get the information for the daughters' daughters in the last days and I just discovered there's such a tool, so I thought it might be useful to other people too.
But I can remove it, if you prefer so ;)

betatim:
The difference between the `TupleTool` and the `DTF_` functors is something
I'd save for later. Basically a dedicated lesson on the `DTF_` functors in
https://github.com/lhcb/second-analysis-steps

13-decay-tree-fitter.md
```diff
@@ -42,8 +42,10 @@ dtt.Dstar.ConsD.daughtersToConstrain = ['D0']
 ```
 Note that you can constrain more than one intermediate state at once if that fits your decay. 
 
+When using the DecayTreeFitter in a DecayTreeTuple, all the variables created by the other TupleTools are not affected by the change, but some new variables are created, one set per DecayTreeFitter instance. Depending on whether the `Verbose` option is specified, the new variables are created for the head particle only or for the head particle and its daughters too. If the daughters are not stable particles and decay further, the daughters of the daughters have no new variables associated to them. In some cases it might be useful to make this information available too. For more information on that, see [this](http://svn.cern.ch/guest/lhcb/Erasmus/trunk/Phys/Bd2D0Kst/DaVinci/TupleBuilder.py) example of how to use the DecayTreeFitter via LoKi functors.
```
betatim: if we keep the link to `TupleBuilder.py` I have two requests: can we make it a link that will work for ever (include a specific revision number maybe)? The fear is that it will stop containing an example and we still link to it.

The other thing is to give it a better name than [this]. Maybe just move the link to be the words "example of how to use DecayTreeFitter via LoKi functors"

FedericaLionetto:
I would still mention that it is possible to add the information for the final state particles too, maybe saying that this involves using the DecayTreeFitter via LoKi functors (no link to TupleBuilder provided) and that the way to do that will be discussed in detail in one of the advanced lectures? I t think it's important that people know this is feasible, just that :)

betatim:
You could make a callout box saying "Daughters of daughters can be done with `DTF_`, see the `second-analysis-steps` for details"

FedericaLionetto:
Agreed! :)

FedericaLionetto:
I modified the text, so let me know if you are happy with this version and what we have to do now ;)

betatim:
Can you make all occurrences of DecayTreeFitter, DecayTreeTuple, etc into `DecayTreeFitter`? Sorry to be a pain in the backside.

Otherwise it looks good to me, so I will merge it once that is done.

FedericaLionetto:
Done! No problem, I might be worse than you for this kind of things :P

betatim:
Thanks for this and congratulations on your first Pull Request :fireworks: :ice_cream:

FedericaLionetto:
Yuppie! :smiley: 
Should I delete the branch now?

betatim:
Uhm, which branch is this? The one in your repository? You can do what you
want with that one, usually I remove them to keep things tidy.

PR #135, opened 2015-07-21
Add note about marking all particles when using branches

ibab:
When using branches, you have to mark all particles you want to specify in the main decay descriptor, otherwise DaVinci will ignore them.

### DIFF
--- a/12-add-tupletools.md
+++ b/12-add-tupletools.md
@@ -53,18 +53,18 @@ The way the `DecayTreeTuple.Decay` is written in in our [minimal DaVinci job](08
 dtt.Decay = '[D*(2010)+ -> (D0 -> K- pi+) pi+]CC'
 ```
 
-means that the configured `TupleTools` will only run on the head of the decay chain, that is, the D*.
-In order to select more particles for which we want the information stored, we need to mark them with a `^` symbol in the decay descriptor.
-For example, if we want to fill the information of the D0 and its children, we would modify the `dtt` to look like this:
+means that the configured `TupleTools` will only run on the head of the decay chain, that is, the D\*.
+In order to select the particles for which we want the information stored, we need to mark them with a `^` symbol in the decay descriptor.
+For example, if we want to fill the information of the D0 and all its children, we would modify the `dtt` to look like this:
 
 ```python
-dtt.Decay = '[D*(2010)+ -> ^(D0 -> ^K- ^pi+) pi+]CC'
+dtt.Decay = '[D*(2010)+ -> ^(D0 -> ^K- ^pi+) ^pi+]CC'
 ```
 
 This will run all the configured `TupleTools` on the marked particles, with the caveat that some tools are only run on certain types of particles (eg, tracking tools on particles that have an associated track).
 This configuration is not optimal, since there may be tools which we only want to run on the D's and some only on the children. Enter `Branches`, which allow us to specify which tools get applied to which particle in the decay (in addition to the `TupleTools` configured at the top level).
 
-Branches let you define custom namespaces in your ntuple by means of a `dict`.
+**Branches** let you define custom namespaces in your ntuple by means of a `dict`.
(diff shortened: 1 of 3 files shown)


apuignav:
The text looks good, I think `code/12-add-tupletools/*.py` should also be updated.

ibab:
Okay, I've updated the code examples.

12-add-tupletools.md
```diff
@@ -53,18 +53,18 @@ The way the `DecayTreeTuple.Decay` is written in in our [minimal DaVinci job](08
 dtt.Decay = '[D*(2010)+ -> (D0 -> K- pi+) pi+]CC'
 ```
 
-means that the configured `TupleTools` will only run on the head of the decay chain, that is, the D*.
+means that the configured `TupleTools` will only run on the head of the decay chain, that is, the D\*.
 In order to select the particles for which we want the information stored, we need to mark them with a `^` symbol in the decay descriptor.
-For example, if we want to fill the information of the D0 and its children, we would modify the `dtt` to look like this:
+For example, if we want to fill the information of the D0 and all its children, we would modify the `dtt` to look like this:
```
alexpearce: When the soft pion wasn't marked this statement was true, now you'll need to change it to “…fill the information of the D*+ and all of it's children, …”.

ibab: Yes, thanks for spotting this!

ibab:
I've opted for

```
`D0` and its children, as well as the soft `pi+`
```

Would this be okay?

12-add-tupletools.md
```diff
@@ -53,18 +53,18 @@ The way the `DecayTreeTuple.Decay` is written in in our [minimal DaVinci job](08
 dtt.Decay = '[D*(2010)+ -> (D0 -> K- pi+) pi+]CC'
 ```
 
-means that the configured `TupleTools` will only run on the head of the decay chain, that is, the D*.
+means that the configured `TupleTools` will only run on the head of the decay chain, that is, the `D*`.
```
alexpearce: Could you change `D*` to `D*(2010)+`? Then all the `code` particles match the decay descriptor.

12-add-tupletools.md
```diff
@@ -53,18 +53,18 @@ The way the `DecayTreeTuple.Decay` is written in in our [minimal DaVinci job](08
...
 ```python
-dtt.Decay = '[D*(2010)+ -> ^(D0 -> ^K- ^pi+) pi+]CC'
+dtt.Decay = '[D*(2010)+ -> ^(D0 -> ^K- ^pi+) ^pi+]CC'
 ```
 
 This will run all the configured `TupleTools` on the marked particles, with the caveat that some tools are only run on certain types of particles (eg, tracking tools on particles that have an associated track).
 This configuration is not optimal, since there may be tools which we only want to run on the D's and some only on the children. Enter `Branches`, which allow us to specify which tools get applied to which particle in the decay (in addition to the `TupleTools` configured at the top level).
 
-Branches let you define custom namespaces in your ntuple by means of a `dict`.
+**Branches** let you define custom namespaces in your ntuple by means of a `dict`.
```
alexpearce: Maybe italics instead of bold?

alexpearce:
Sorry for taking so long to read this…

alexpearce:
@ibab ping :satellite:

ibab:
Good suggestions!
I'll implement them later today.

ibab:
Okay, done, even if though it's not the same day

alexpearce:
Gah now there are conflicts as #142 has been merged. Sorry about that. Would you mind rebasing? I can do it if you're busy, just let me know.

alexpearce:
Are you able to fix the conflicts @ibab?

ibab:
Okay, I've rebased the changes

alexpearce:
LGTM, thanks!

PR #136, opened 2015-08-04
Remove mention of lhcbstarterkit8 from the prerequisits

betatim:
We keep getting emails from people asking where the machine is. I don't think anything special (beyond `lxplus`) is needed for the starterkit as it stands now.


alexpearce:
Thank you!

kdungs:
Yes, thank you. :cake:

PR #142, opened 2015-11-07
A few suggestions for the DecayTreeTuple lesson

pieterdavid:
- doxygen provides a nice list of TupleTools (one for each type). I think that is more convenient than clicking through the directory structure in the svn web interface (being able to figure out where the code is in svn is also very useful, but rather belongs in a separate lesson about debugging and development).
- there should never be a need to mark the head of a decay descriptor, so I would remove that ^ in the example (and make the wording a little bit more explicit about the distinction).

### DIFF
--- a/12-add-tupletools.md
+++ b/12-add-tupletools.md
@@ -55,7 +55,7 @@ dtt.Decay = '[D*(2010)+ -> (D0 -> K- pi+) pi+]CC'
 
 means that the configured `TupleTools` will only run on the head of the decay chain, that is, the D*.
 In order to select more particles for which we want the information stored, we need to mark them with a `^` symbol in the decay descriptor.
-For example, if we want to also fill the information of the D0 and its children, we would modify the `dtt` to look like this:
+For example, if we want to fill the information of the D0 and its children, we would modify the `dtt` to look like this:
 
 ```python
 dtt.Decay = '[D*(2010)+ -> ^(D0 -> ^K- ^pi+) pi+]CC'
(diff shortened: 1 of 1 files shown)


12-add-tupletools.md
```diff
@@ -54,8 +54,8 @@ dtt.Decay = '[D*(2010)+ -> (D0 -> K- pi+) pi+]CC'
 ```
 
 means that the configured `TupleTools` will only run on the head of the decay chain, that is, the D*.
-In order to select the particles for which we want the information stored, we need to mark them with a `^` symbol in the decay descriptor.
-For example, if we want to fill the information of the D0 and its children, we would modify the `dtt` to look like this:
+In order to select more particles for which we want the information stored, we need to mark them with a `^` symbol in the decay descriptor.
+For example, if we want to also fill the information of the D0 and its children, we would modify the `dtt` to look like this:
```
betatim: I'd remove of the 'also' again.

12-add-tupletools.md
```diff
@@ -87,22 +87,10 @@ The usage of `Branches` is very important (and strongly encouraged) to keep the
...
-> - `DecayTreeTupleANNPID` for the NeuralNet-based PID tools.
-> - `DecayTreeTupleDalitz` for Dalitz analysis.
-> - `DecayTreeTupleJets` for obtaining information on jets.
-> - `DecayTreeTupleMC` gives us access to MC-level information.
-> - `DecayTreeTupleMuonCalib` for muon calibration tools.
-> - `DecayTreeTupleReco` for reconstruction-level information, such as `TupleToolTrackInfo`.
-> - `DecayTreeTupleTracking` for more detailed tools regarding tracking.
-> - `DecayTreeTupleTrigger` for accessing to the trigger information of the candidates.
+> `TupleTools` are Gaudi tools that implement the [`IParticleTupleTool`](http://lhcb-release-area.web.cern.ch/LHCb-release-area/DOC/davinci/releases/latest/doxygen/d1/d77/class_i_particle_tuple_tool.html) interface (if they fill information about an `LHCb::Particle`) or the [`IEventTupleTool`](http://lhcb-release-area.web.cern.ch/LHCb-release-area/DOC/davinci/releases/latest/doxygen/d7/d5c/class_i_event_tuple_tool.html) interface (if they fill information about the event in general), so we can get the complete list of `TupleTools` of each type at the top of the doxygen page of the interface.
```
betatim: Instead of getting rid of the existing lines I would add this after. For `TupleTools` we can link to doxygen but getting people used to looking directly in svn is IMO very valuable.

We might need to add a callout box explaining in very brief terms what interfaces, Gaudi tools, etc are as none of these concepts have been explained yet.

betatim:
Thanks for this.

At least part of pointing directly to the svn is to get people using it (even if in this case doxygen contains something useful).

pieterdavid:
I implemented your comments. Pointing people to svn to get them used to it is also useful indeed.

betatim:
Thank you, adding :ice_cream: to the offerings from #mozfest

PR #143, opened 2015-11-07
Remote access with XML catalog as alternative for downloading DSTs

pieterdavid:
Slightly more advanced than downloading files, but I think it is good to know that that can relatively easily be avoided

### DIFF
--- a/04-files-from-grid.md
+++ b/04-files-from-grid.md
@@ -46,6 +46,22 @@ Again this will take a while but afterwards you should have a file called `00035
 
 Since those tuples tend to be quite large, you might want to use your AFS work directory instead of your AFS user directory to store files.
 
+> ## Alternative: read the remote files instead of downloading them {.callout}
+> To avoid filling up your AFS quota with dst files, you can also pass Gaudi an XML catalog such that it can access them remotely.
+>
+> First generate the XML catalog with
+> ```bash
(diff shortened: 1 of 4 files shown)


betatim:
Thanks for this!

What do others think about adding in the appropriate `code/` directory an example that uses the XML catalog instead of the local file?

ibab:
:+1: on `code/` example

pieterdavid:
Something like this?

04-files-from-grid.md
```diff
@@ -46,6 +46,22 @@ Again this will take a while but afterwards you should have a file called `00035
 
 Since those tuples tend to be quite large, you might want to use your AFS work directory instead of your AFS user directory to store files.
 
+> ## Alternative: read the remote files instead of downloading them {.callout}
```
betatim: "read the remote files" -> "read files remotely instead of"

betatim:
Looks good to me (just fix the little english language thing), then we can merge.

We send :cake: from #MozFest. Thanks

04-files-from-grid.md
```diff
@@ -46,6 +46,22 @@ Again this will take a while but afterwards you should have a file called `00035
 
 Since those tuples tend to be quite large, you might want to use your AFS work directory instead of your AFS user directory to store files.
 
+> ## Alternative: read files remotely instead of downloading them {.callout}
+> To avoid filling up your AFS quota with dst files, you can also pass Gaudi an XML catalog such that it can access them remotely.
```
alexpearce: Uppercase DST.

alexpearce:
This is great, thank you! And sorry so the enormous delay 🙇 There's just one tiny change, and then I'd so it's ready to merge.

pieterdavid:
In the mean time I can't even test this anymore, so I just changed the one word and hope the rest is still fine :-)

alexpearce:
Let's hope so 😉

PR #145, opened 2015-11-09
Add slightly information on EOS

dominikmuller:
Added a comment on the EOS_HOME variable and mentioned at the bottom that EOS paths can be used as the target of hadd calls.

### DIFF
--- a/09-eos-storage.md
+++ b/09-eos-storage.md
@@ -38,7 +38,8 @@ MassStorageFile = {'fileExtensions': [''],
 
 The line should look very similar to this already, the only thing that
 needs changing is the `path` entry. You should change it to your EOS home directory
-which is `/eos/lhcb/user/a/another` if your username is `another`.
+which is `/eos/lhcb/user/a/another` if your username is `another`. It is also stored
+in the `EOS_HOME` environment variable.
 
 > ## Automatically transfer files {.callout}
 >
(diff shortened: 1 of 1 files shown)


alexpearce:
LGTM :+1: and thanks very much for your first contribution!

betatim:
Does it make sense to add some words about `xrdfs` and `xrdcopy`? If not, :+1: on merge

(Being greedy with @dominikmuller's time :smile:)

:cake: for Dominik for his first contribution

alexpearce:
Indulge my laziness: what does `xrdfs` do? I've only used `xrdcp` from the `xrd*` family.

betatim:
It is like a shell for poking around xrootd filesystems. More `ls` and `rm` than copy and move. `xrdcp` could be useful for moving files that aren't `root` files.

alexpearce:
Hmm maybe then. I guess it's 'more recommended' than `eosmount`?

dominikmuller:
We could mention the two commands, though I have only used the interactive EOS shell for most things.

betatim:
`eosmount` is a good point. I learnt about the `xrd*` family because I wanted to use EOS without FUSE. If you don't care about that `eosmount` is easier (if you are on lxplus).

alexpearce:
There's a conflict preventing the merge @dominikmuller.

I suggest opening an issue/another PR for adding more information.

eduardo-rodrigues:
Better close this tiny update that's been open for 6.5 years!

PR #148, opened 2015-11-13
Replaced MM functor with M functor

KonstantinSchubert:
We were told during the lesson that the MM mass gives the combined mass without any topological constraints and that we should thus use the M mass. 

I suppose this translates to the functors? The twiki is not very clear here. In this case, we should be using the M functors instead of the MM functors. I tried to replace them accordingly in this lesson

### DIFF
--- a/12-add-tupletools.md
+++ b/12-add-tupletools.md
@@ -144,8 +144,8 @@ To add LoKi-based leaves to the tree, we need to use the `LoKi::Hybrid::TupleToo
   - The `Variables` property, consisting of a `dict` of (variable name, LoKi functor) pairs. In here, LoKi functors can be used, as well as any variable we may have defined in the `Preambulo`:
     ```python
     dstar_hybrid.Variables = {
-        'mass': 'MM',
-        'mass_D0': 'CHILD(MM, 1)',
+        'mass': 'M',
+        'mass_D0': 'CHILD(M, 1)',
         'pt': 'PT',
         'dz': 'DZ',
         'dira': 'BPVDIRA',
@@ -155,7 +155,7 @@ To add LoKi-based leaves to the tree, we need to use the `LoKi::Hybrid::TupleToo
         'n_highpt_tracks': 'NINTREE(ISBASIC & HASTRACK & (PT > 1500*MeV))'
     }
     d0_hybrid.Variables = {
-        'mass': 'MM',
+        'mass': 'M',
         'pt': 'PT',
         'dira': 'BPVDIRA',
         'vtx_chi2': 'VFASPF(VCHI2)',
--- a/code/12-add-tupletools/ntuple_options_loki.py
+++ b/code/12-add-tupletools/ntuple_options_loki.py
@@ -35,8 +35,8 @@ dstar_hybrid.Preambulo = preamble
 d0_hybrid.Preambulo = preamble
 
 dstar_hybrid.Variables = {
-    'mass': 'MM',
-    'mass_D0': 'CHILD(MM, 1)',
+    'mass': 'M',
+    'mass_D0': 'CHILD(M, 1)',
     'pt': 'PT',
     'dz': 'DZ',
     'dira': 'BPVDIRA',
@@ -45,7 +45,7 @@ dstar_hybrid.Variables = {
     'n_highpt_tracks': 'NINTREE(ISBASIC & HASTRACK & (PT > 1500*MeV))'
 }
 d0_hybrid.Variables = {
-    'mass': 'MM',
+    'mass': 'M',
     'pt': 'PT',
     'dira': 'BPVDIRA',
     'vtx_chi2': 'VFASPF(VCHI2)',


alexpearce:
Awesome, thanks for the PR Konstantin!

I agree with the logic. `MM` returns the mass using the sum of the four-vectors of the children as the ‘parent’, whereas `M` uses the result of the vertex fit, which is allowed to tweak the child momentum, and hence change the mass slightly.

betatim:
As this question keeps popping up all over LHCb and is usually answered with "ergh, maybe or maybe not, who knows, probably doesn't matter" can we add a callout box to this lesson with the explanation you just gave on `MM` vs `M`?

roelaaij:
There was recently a rather long email thread about this, which we should dig up and check that my explanation is correct. I think a callout box is a good idea.

seneubert:
Which one you use depends very much on what you want to do. So "use M" is
too simplified instruction.
Adding info that explains the difference is a good idea.
Check that M actually changes when you do the kinematic fit (I think it
doesn't).
If you want the refitted 4-momenta you have to use the dtf functors.
If you want a much better thing there is a development branch where I
implemented a tool that allows you to run any functor on a refitted tree
(dtf functors will repeat the fit each time you ask for another variable)

On Fri, Nov 13, 2015 at 11:59 AM, Tim Head notifications@github.com wrote:

> As this question keeps popping up all over LHCb and is usually answered
> with "ergh, maybe or maybe not, who knows, probably doesn't matter" can we
> add a callout box to this lesson with the explanation you just gave on MM
> vs M?
> 
> —
> https://github.com/lhcb/first-analysis-steps/pull/148#issuecomment-156397532
> .

## 

---

Dr. Sebastian Neubert

## University Heidelberg

roelaaij:
Quoting Vanya from the mail thread below, which I think this could be used as a basis for the callout box. I agree with Vanya that either M or DecayTreeFitter should be used.

Quote:

Let me confirm that in LoKi::VertexFitter,     the data field LHCb::Particle::m_measuredMass  ( used  in factor "MM")  is calculated  from sum of 4-momenta of all daughter particles extrapolated to the mother vertex. In some sense it is the most precise pure kinematical definition of mass WIHTOUT constraints (e.g. vertex...).

You see - it will be a bit different from LHCb::Particle::m_momentum.M()  ( due to vertex constrains) and it will be  a bit different  from blind sum of 4-momenta of daughter particles (due to extrapolation of daughters).

Honestly speaking I personally do not understand why people are using MM at all...  - it could be useful for unndertsanding of vertex constrains, but fro all physics purposed, clearly  only "M" should be used  (or , much better, mass  after full DTF-refit)

alexpearce:
It depends on the decay as to whether the two are different, I think.

[Here's the thread](https://groups.cern.ch/group/lhcb-hlt2-development/Lists/Archive/Flat.aspx?RootFolder=%2fgroup%2flhcb%2dhlt2%2ddevelopment%2fLists%2fArchive%2fdifference%20between%20XX%5fMM%20and%20XX%5fM&FolderCTID=0x012002002E0CEEFA9819C44693EE27ECFD93D17B) from `lhcb-hlt2-development`, and here's Vanya's reply:

> Let me confirm that in LoKi::VertexFitter,     the data field LHCb::Particle::m_measuredMass  ( used  in factor "MM")  is calculated  from sum of 4-momenta of all daughter particles extrapolated to the mother vertex.
> In some sense it is the most precise pure kinematical definition of mass WIHTOUT constraints (e.g. vertex...).
> You see - it will be a bit different from LHCb::Particle::m_momentum.M()  ( due to vertex constrains) and it will be  a bit different  from blind sum of 4-momenta of daughter particles (due to extrapolation of daughters).
> 
> Honestly speaking I personally do not understand why people are using MM at all...  - it could be useful for unndertsanding of vertex constrains, but fro all physics purposed, clearly  only "M" should be used  (or , much better, mass  after full DTF-refit) 

[And here's another, this time from 2009](https://groups.cern.ch/group/lhcb-davinci/Lists/Archive/Flat.aspx?RootFolder=%2Fgroup%2Flhcb%2Ddavinci%2FLists%2FArchive%2FMM%20%28%20and%20%5FMM%29%20versus%20M%20%20CombineParticlesFilterDesktop%20%28and%20indirectly%20DecayTreeTuple%29&FolderCTID=0x01200200B795ECDBD8943F4E90E5F02A017196C9).

alexpearce:
@roelaaij beat me :wink:

apuignav:
Should we merge this?

alexpearce:
:+1:

betatim:
@KonstantinSchubert do you want to add a callout box as part of this PR? If not then we should create an issue and merge this.

KonstantinSchubert:
Honestly I am stil confused about the issue.

Let me try to summarize:

MM  is LHCb::Particle::m_measuredMass which is  calculated from sum of 4-momenta of all daughter particles extrapolated to the mother vertex. (What does "xtrapolated to the mother vertex" mean?)

M is LHCb::Particle::m_momentum.M(), which is calculated using the 4-momenta of the daugther particles and additonal vertex constraints (which exactly?)

And then there is the mass from the DecayTreeFitter, which additionally includes the constraints applied in DecayTreeFitter.

I am fine if we merge without adding a callout box, I am also willing to try adding a callout box if I get a bit of hand-holding with the formulation.

alexpearce:
I don't think that's quite right.

`MM` is computed using the sum of four momenta of the daughters, as measured by the track reconstruction. `M` is computed using the output of the _vertex fit_, which is allowed to wiggle the daughter momenta within their uncertainties in order to bring them closer to a singe (x, y, z) point (to match the hypothesis that they originate from the same, physical particle).

DecayTreeFitter essentially reconstructs the vertices by fitting the entire decay at once, whereas the default LHCb reconstruction creates vertices by combining objects leaf-by-leaf. Taking the D*\+ decay as an example, in the default LHCb reconstruction the creation of the D*+ vertex does not change the properties of the D0 vertex. With DecayTreeFitter, all input parameters can be varied simultaneously.

saschastahl:
I think we should merge this.

alexpearce:
We should get this merged ASAP. @KonstantinSchubert: would you mind also changing [`13-decay-tree-fitter.md`](https://github.com/lhcb/first-analysis-steps/blob/4fe24af53f25f7b8c90a8b939b1555f8aa235f31/13-decay-tree-fitter.md) and [`ntuple_options_loki.py`](https://github.com/lhcb/first-analysis-steps/blob/826b492c91b6b2c36a16487f80704ce5c7c34057/code/12-add-tupletools/ntuple_options_loki.py#L37-L53) to change `MM` to `M`?

KonstantinSchubert:
Maybe in  13-decay-tree-fitter.md it really makes some sense to use MM? Or at least this is maybe a good place for the callout box?

alexpearce:
Hmm yes, you're right. The text calls it the "raw mass", which in some sense it is. This does look like one place where the callout might be good.

saschastahl:
Agreed.

alexpearce:
I'm going to merge this, we can add a callout in another PR.

PR #149, opened 2015-11-13
Swap order of lessons 10 and 11.

roelaaij:
These lessons were in the wrong order, so swap them. Also rename the eos lesson file to have the right chapter number. Also some trailing whitespace.


alexpearce:
LGTM :+1: Thank you!

PR #150, opened 2015-11-13
Fix dtt.Decay

mick-mulder:
Fixed mistake in dtt.Decay definition such that not only Dstar variables, but all variables for all particles, are written to ntuple.

### DIFF
--- a/code/12-add-tupletools/ntuple_options.py
+++ b/code/12-add-tupletools/ntuple_options.py
@@ -9,7 +9,7 @@ line = 'D2hhCompleteEventPromptDst2D2RSLine'
 # Create an ntuple to capture D*+ decays from the StrippingLine line
 dtt = DecayTreeTuple('TupleDstToD0pi_D0ToKpi')
 dtt.Inputs = ['/Event/{0}/Phys/{1}/Particles'.format(stream, line)]
-dtt.Decay = '[D*(2010)+ -> (D0 -> K- pi+) pi+]CC'
+dtt.Decay = '[D*(2010)+ -> ^(D0 -> ^K- ^pi+) ^pi+]CC'
 track_tool = dtt.addTupleTool('TupleToolTrackInfo')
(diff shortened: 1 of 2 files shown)


betatim:
I think it is like this on purpose as the corresponding lesson goes on to tell people that probably the best thing to do is use explicit branches. What do the others thing?

If we change it we should probably also change `code/12-add-tupletools/ntuple_options_loki.py`

betatim:
I forgot: Welcome to contributing to the starterkit!

mick-mulder:
Thank you! 

Indeed, this lesson advises people to use branches. However, the branches with daughter particles cannot be filled unless the corresponding daughter particles have been marked with a ^ in the dtt.Decay definition. After copying this file directly, I got the following error message (for all of the daughters, I have just included this one for brevity):

> TupleDstToD0pi_...SUCCESS  #ERRORS     = 1        Message = 'The decay descriptor 'TupleDstToD0pi_D0ToKpi.D0 : ( ( D_(2010)+  -> ^( ( D0  -> K-  pi+ ) ) pi+ ) || ( D_(2010)-  -> ^( ( D~0  -> K+  pi- ) ) pi- ) ) ' returned a match which is not also matched by your main decay descriptor. Ignoring it.'

As a result, the variables were not written to the ntuple. 

Including the marks in the decay description made the code run without these errors, and led to an ntuple with the variables associated with the daughter particles.

betatim:
Ah ok, can you also change it in `code/12-add-tupletools/ntuple_options_loki.py` by pushing another commit to the same branch in your fork. Then we can merge it.

mick-mulder:
I will change it right now, thanks for the heads-up!

mick-mulder:
I just included the same fix for `code/12-add-tupletools/ntuple_options_loki.py`, could the branches now be merged?

betatim:
Purfect :cat2:

Thank you for contributing: :beers:

PR #167, opened 2016-03-04
fixed a misleading information in TupleToolDecayTreeFitter use

cpviolation:
Hi everybody,

I've noticed that there is misleading information in tupletooldecaytreefitter usage description.
It is written that LoKi functors are needed to access daughters information when there is a method to do that in the tuple tool since 3 years.

The real changes I propose are in L47, the rest is just my editor unnecessarily beautifying the code.

Maurizio Martinelli

### DIFF
--- a/13-decay-tree-fitter.md
+++ b/13-decay-tree-fitter.md
@@ -1,7 +1,7 @@
 ---
 layout: page
 title: HowTo DecayTreeFitter
-subtitle: How do I use DecayTreeFitter? 
+subtitle: How do I use DecayTreeFitter?
 minutes: 10
 ---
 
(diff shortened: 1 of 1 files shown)


betatim:
Looks fine to me. Especially as the link doesn't go to a solution.

Does someone remember what the original motivation for writing this was?

(Does one of the other starterkit people know why the lines that weren't changed seem to differ in the last empty char? Is this win vs mac vs linux line endings?)

seneubert:
Oh wow, I did not know this. Actually very few people seem to know. Very useful indeed.
Also makes my implementation of the DTFDict tool, (for which I have been requesting feedback on the DaVinci mailing list) less needed.

Where will the updated Daughters appear?

cpviolation:
It will create branches of the type:
`Dst_ReFit_D0_piplus_1_PX`

seneubert:
I am a bit disappointed by this feature.
In my example Lb2LcD0K I contrain all three intermediates (including Lb) but UpdateDaughters only gives me updated momenta for the final state daughters, not for the Lc and D0 ...
not super serious, but annoying.

The merge request is good, though

cpviolation:
The intermediate particles momenta are not saved since they will be the same as the sum of the daughters.
This is done to avoid duplicating the information and producing very large datasets at the early stage of the analysis (typically the selection).

seneubert:
Obviously you can do this. 
But that is not how most people use TupleTools. Actually that is contrary to the whole philosophy of the tupletools as they are being used.
I don't say that it doesn't make sense, just saying that it's kind of unintuitive.

cpviolation:
Actually, I sticked to the philosophy of TupleToolDecayTreeFitter, where only the updated mass and absolute momentum is retained after the fit.
Anyway, discussing about what would be the best option is a matter of taste, and I think that it is not the right thread to discuss it.

seneubert:
Being able to see the refitted daughters is important for some analyses.
Whenever you do an amplitude analysis you need this.

betatim:
Can this be merged? Yes, no, maybe?

alexpearce:
Oh yeah 👍 

Thanks @cpviolation!

betatim:
🎉 !

PR #170, opened 2016-06-01
Added reference to LHCb Q&A

KonstantinSchubert:
As discussed in this issue https://github.com/lhcb/starterkit/issues/21 (which was unfortunately posted to the wrong projet.)

I added the reference after the mention of the mailing lists since LHCb Q&A is supposed to at least partially replace the mailing lists.So this seemed like the most natural place to add this.
If you think this is too prominently placed, I can of course change it.


alexpearce:
LGTM 👍

KonstantinSchubert:
Thank you all!

Albert Puig notifications@github.com schrieb am Mi., 1. Juni 2016 17:40:

PR #176, opened 2016-06-15
Adding bash as default

henryiii:
Added updated information on Bash being the default, but still includes the method to change.


alexpearce:
LGTM.

Issue #175, opened 2016-06-15
File number


henryiii:
The file used seems to change from `00035742_00000001_1.allstreams.dst` to `00035742_00000002_1.allstreams.dst` randomly. If using the fast download of one file (the 1 file), the other file (2) does not work. Would it be better to use just the first file when only one file is used?

alexpearce:
Yes, we should be consistent with the file we use.

ibab:
This has already been fixed by #177

PR #177, opened 2016-06-15
Using the 000001_1 file everywhere, instead of sometimes using 000002_1

henryiii:
This uses just the one file (that ends in `1_1`) everywhere that a single file is used. This is the file suggested at the beginning to download if the system is slow. (If all the files are downloaded at the beginning, it does not make a large difference, other than being more consistent).

### DIFF
--- a/04-files-from-grid.md
+++ b/04-files-from-grid.md
@@ -43,17 +43,17 @@ Initialisation of the proxy might take a while and should ask you for your certi
 Once we have a working Dirac installation, getting the file is as easy as
 
 ```bash
-dirac-dms-get-file LFN:/lhcb/MC/2012/ALLSTREAMS.DST/00035742/0000/00035742_00000001_1.allstreams.dst
+dirac-dms-get-file LFN:/lhcb/MC/2012/ALLSTREAMS.DST/00035742/0000/00035742_00000002_1.allstreams.dst
 ```
 
-Again this will take a while but afterwards you should have a file called `00035742_00000001_1.allstreams.dst` in the directory where you called the command.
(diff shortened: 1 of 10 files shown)


henryiii:
PS: Closes #175.

henryiii:
Note: this does affect some of the plots, so I could use `2_1` everywhere instead. The first plot on the DecayTreeFitter page has one outlier in `1_1`.

c1

henryiii:
The 002_1 file is not available with 

``` bash
xrdcp root://eoslhcb.cern.ch//eos/lhcb/user/a/apearce/Starterkit/Nov2015/00035742_00000002_1.allstreams.dst .
```

alexpearce:
I hope you didn't spend too long being confused on this 😄  Thanks for fixing it.

> The 002_1 file is not available with
> 
> ```
> xrdcp root://eoslhcb.cern.ch//eos/lhcb/user/a/apearce/Starterkit/Nov2015/00035742_00000002_1.allstreams.dst
> ```

Will that be a problem after this PR?

henryiii:
Can you put the 0000002_1 file in your eos directory? Then I can replace the 00001_1's with 00002_1, and the plots will match the ones in the starterkit.

alexpearce:
Sure thing, it's there now.

```
> eos ls /eos/lhcb/user/a/apearce/Starterkit/Nov2015/
00035742_00000001_1.allstreams.dst
00035742_00000002_1.allstreams.dst
```

henryiii:
Okay, the PR should be more consistent now with previous graphs. Thanks! 🙇

betatim:
Can this be merged?

alexpearce:
:+1:

Thanks a lot @henryiii 👏

betatim:
🍰 🍰 🍰

PR #178, opened 2016-06-21
Latest davinci

henryiii:
This bring DaVinci up to a much newer version. Everything seemed to work properly with the new version (save for needed fix in #172, but that has identical behavior on both versions)


alexpearce:
I can't see any remaining instances of `v36r6`, so I think this is good to go.

Keeping up with recent DV versions is something we could try to automate at some point.

PR #183, opened 2016-10-11
Few corrections for datamanagement

chaen:


04-files-from-grid.md
```diff
@@ -34,8 +34,8 @@ quick look at the dataset, it is sufficient to just obtain one of those files.
 
 Before we can download the file, we need to set up our connection with the grid and load the Dirac software:
 ```bash
-lhcb-proxy-init
 lb-run LHCbDirac $SHELL
+lhcb-proxy-init
```
alexpearce: Maybe we should just change this to:

``` shell
lb-run LHCbDirac lhcb-proxy-init
```

11-eos-storage.md
```diff
@@ -53,6 +53,9 @@ element on the Grid. For files which you do not plan to work with
 interactively and instead feed into a different grid job it makes
 sense to leave them on the grid.
 
+Note that EOS *IS* a grid storage (the CERN one), but your personnal space is just
+in a different subdirectory than the files "on the grid" managed by DIRAC.
+
```
alexpearce, line 57: This is a nice addition, it should clarify some things. I would re-word slightly:

> Note that EOS _is_ 'Grid storage', specifically that which is located at CERN. Your personal space, at `/eos/lhcb/user/<your username's first letter>/<your username>`, is just another location on EOS, and happens to be different to the default location used by DIRAC for 'grid storage'.

11-eos-storage.md
```diff
@@ -113,6 +116,9 @@ Once your job has completed and the files have been copied to
 EOS by `ganga` you can access them from your terminal by mounting
 your EOS area.
 
+Important: mounting eos is discouraged by the LHCb data management team,
+and no support will be provided by them for problems/questions related to it.
+
```
alexpearce, line 121: I think this should be in a callout box, to highlight it. Is there a source we could link to for this?

04-files-from-grid.md
chaen: Why not, but then bellow the lb-run should be added again for the dirac-dms-get-file. 
Do you prefer it this way ?

11-eos-storage.md
chaen: yep, better

11-eos-storage.md
chaen: What do you mean a source ? Where we state that ? I would have to dig for it

04-files-from-grid.md
alexpearce: Good point. I don't have a _strong_ preference, but I like that using `lb-run` makes it very obvious what's happening.

11-eos-storage.md
alexpearce: OK, that's no problem then. I wondered if there was something obvious, but we say lots of things should or shouldn't be done without giving a reason. This page will become the source 😄

alexpearce:
Thanks a lot for the PR! It's very helpful. If you could just add `lb-run LHCbDirac` to the other commands, I think we're good to go.

11-eos-storage.md
chaen: Actually it was discussed in some presentations, maybe LHCbWeek somewhere. There are plenty of technical reasons, but well.... :-)

chaen:
I am fine adding the lb-run in front of the other commands, but I would like that it is stated somewhere clearly (not necessarily in this page) that one could avoid it by starting a new bash once. Maybe linking the documentation of lb-run is the best thing to do

alexpearce:
That's a good point. I think we should clarify what `lb-run` does, but that would be better suited to another lesson. I'll open an issue for it.

alexpearce:
Thank you very much for the PR! It's much clearer now. 🍰 🍨

PR #190, opened 2016-10-31
	EOS can be used on lxplus without mounting it

closier:


ibab:
This looks good to me!
I wasn't aware of the fact that eos is already mounted.

Issue #182, opened 2016-10-11
update Developing LHCb Software to use git


pikacic:
All major LHCb projects migrated to Git this year, so it's time to merge https://github.com/lhcb/second-analysis-steps/blob/master/02-lb-git.md in `15-lhcb-dev.md` (and remove the git lesson from second-analysis-steps).

pikacic:
I could prepare a pr some times next week.

alexpearce:
I wonder if it's worth teaching this stuff during the git lessons. Usually we don't teach anything LHCb specific in the [first couple of days](http://lhcb.github.io/analysis-essentials/), but maybe it makes sense there.

PR #197, opened 2016-11-25
update Developing LHCb Software to use git

pikacic:
See #182

### DIFF
--- a/davinci-grid.md
+++ b/davinci-grid.md
@@ -123,7 +123,7 @@ To look at the `root` file produced by the job start a new terminal, and
 type:
 
 ```bash
-$ lb-run DaVinci v41r2 $SHELL
+$ lb-run DaVinci/v41r2 $SHELL
 $ root -l path/to/the/job/output
 ```
 
--- a/davinci.md
+++ b/davinci.md
@@ -58,7 +58,7 @@ One of the most important ones is *DaVinci*, which provides lots of *Algorithms*
 
 You can run DaVinci using the following command:
 ```bash
-lb-run DaVinci v41r2 gaudirun.py
+lb-run DaVinci/v41r2 gaudirun.py
 ```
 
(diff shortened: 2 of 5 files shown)


05-interactive-dst.md
```diff
@@ -192,10 +192,10 @@ functors](06-loki-functors.html)!
 > For example, to explore the `DST` we could have simply done:
 >
 > ```
-> lb-run Bender bender 00035742_00000002_1.allstreams.dst
+> lb-run Bender/latest bender 00035742_00000002_1.allstreams.dst
```
alexpearce: Why is this `/latest`, but when specifying a version it's with a space?

pikacic: Actually `Project/version` works always (I forgot to update the --help message when I added the feature to `lb-run`).

The syntax with a space works only for "standard" versions (vXrY, latest), but not with more original versions (master, upgradeTracking) that you may find in the nightly builds.

At some point I'll drop the syntax with a space... in a far future.

About this specific line, there's a difference between `lb-run` and  `SetupProject`: when you do not specify the version `SetupProject` uses _latest_, while `lb-run` uses _prod_ which is almost never defined (in which case `lb-run` falls back on `SetupProject`). What is really wanted here is `lb-run Bender/latest <command>` and not `lb-run Bender <command>`.

One more note: LHCbDirac does have a defined _prod_ version, so `lb-run LHCbDirac <command>` works correctly.

alexpearce: Do you think it would be better to always use `Project/version` then?

pikacic:
I just rebased on the the current master.

I thought of merging stuff from `second-analysis-steps/02-lb-git.md`, but I'm not that convinced anymore. This is a lightweight introduction to satellite projects, and the other is more targeted to developers.

ibab:
Yes, I agree with that.
Everything looks good, so I'm merging now.

PR #199, opened 2016-11-25
Add callout box describing setDescriptorTemplate()

olupton:
We recently added a helper function `setDescriptorTemplate()` to `DecayTreeTuple`. It's not released yet, but can easily be added to older versions. This adds a callout box describing how to replace all the `dtt.Decay` and `dtt.addBranches()` stuff with a single line.


ibab:
This looks good!

PR #204, opened 2016-11-27
Remove numbers and update index

ibab:
I've removed the numbers from the lessons and code so that we can reorder them as we want.
This required changing the validation scripts a bit.
I've then updated the Index with the order that we agreed on.
There are still a few lessons missing that we thought about adding:
 - Using DaVinci from Ganga locally (could potentially be done on the fly during the Ganga lesson)
 - Merging files, resubmitting Ganga jobs
 - Using Ganga with custom projects (GaudiExec)


apuignav:
LGTM!

PR #206, opened 2016-11-27
Some extra Ganga info

mesmith75:
This is some ganga info for the data access - introduces BKQuery, DiracFile, accessURL and the gangarc config to set the gangadir.

### DIFF
--- a/ganga-data.md
+++ b/ganga-data.md
@@ -35,7 +35,7 @@ This is a list of `DiracFile`, the Ganga object for files stored on the grid. We
 Ganga In [5]: j.inputdata[0].accessURL()
 Ganga Out [5]: ['root://bw32-4.grid.sara.nl:1094/pnfs/grid.sara.nl/data/lhcb/LHCb/Collision12/SEMILEPTONIC.DST/00051179/0000/00051179_00006978_1.semileptonic.dst']
 ```
-The returned path can be used by Bender to explore the contents of the DST.
+The returned path can be used by Bender to explore the contents of the DST, as in the [Interactively exploring a DST](interactive-dst.html) lesson.
 
 In the previous lesson we looked at the location of the ouput with `job(782).outputdir`. This location points us to the `gangadir` where ganga stores information about the jobs and the output. If we have lots of jobs with large files the file system where the gangadir is located will quickly fill up.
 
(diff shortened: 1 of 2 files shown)


ganga-data.md
```diff
@@ -0,0 +1,56 @@
...
+> ## Setting the gangadir location {.callout}
+>
+> The location of the `gangadir` can be changed in the configuration file '~/.gangarc'. Just search for the `gangadir` attribute and change it to where you like (on the CERN AFS the `work` area is a popular choice).
+
+To avoid filling up the filespace it is wise to put the large files produced by your job somewhere with lots of storge - the grid. You can do so by setting the `outputfiles` attribute:
+
+```python
+j.outputfiles = [DiracFile('*.root'), LocalFile('stdout')]
+```
+The `DiracFile` will be stored in your user area on the grid (with up to 2TB personal capacity) from where you can access it with the `accessURL()` function as before. The wildcard means that any root file produced by your job will stay on the grid. `LocalFile` downloads the file to your `gangadir`, in this case the one called `stdout`. In general you are encouraged to keep your large files on the grid to avoid moving large amounts of data around through your work area.
```
alexpearce: I think there's a default list of files that are always downloaded locally. Is it worth mentioning those, so users don't add them redundantly to their job configuration?

ganga-data.md
```diff
@@ -0,0 +1,56 @@
...
+     localDir = 
+   ) 
+ )
+```
+This is a list of `DiracFile`, the Ganga object for files stored on the grid. We can access one locally via the `accessURL`:
+```python
+Ganga In [5]: j.inputdata[0].accessURL()
+Ganga Out [5]: ['root://bw32-4.grid.sara.nl:1094/pnfs/grid.sara.nl/data/lhcb/LHCb/Collision12/SEMILEPTONIC.DST/00051179/0000/00051179_00006978_1.semileptonic.dst']
+```
+The returned path can be used by Bender to explore the contents of the DST.
```
alexpearce: Could mention the DST explorer lesson too, which I assume will also work with these URLs.

ganga-data.md
mesmith75: Yes - here I just wanted to illustrate what LocalFile means. I'll add in the default list.

ganga-data.md
mesmith75: Ok, I'll put a link in.

alexpearce:
Excellent, thanks a lot!

This should be added to the table of contents on the index page. Where do you think would be the best place for it?

mesmith75:
I think just before the EOS lesson would be best?

alexpearce:
Make sense to me. Do you mind adding a link? Then it should be ready to merge.

alexpearce:
Super, thank you!

PR #207, opened 2016-11-28
Add note about dst-dump at the end of the interactive DST lesson

olupton:
`dst-dump` is a very useful tool, I thought it was worth mentioning here.


alexpearce:
Looks bon to me 🇫🇷

PR #208, opened 2016-11-28
Add callout box about combining LoKi functors

olupton:
There is a non-trivial gotcha with the combination of LoKi functors, which can lead to very strange bugs. This MR adds a callout box with a few words of explanation.


apuignav:
Good point. I usually say it in the session, but it should be written. LGTM.

Issue #203, opened 2016-11-26
Use GaudiExec in all Ganga lessons


cmarinbe:
This is the current recommendation so we should update the lessons to use this. We have been discussing, though, about the necessity to create a `lb-dev` folder even if you want to run from a released version. In principle, `prepareGaudiExec` should create it for you. We need to figure out how this exactly works.

cmarinbe:
I've tried to do: 

```
j = Job()
myApp = prepareGaudiExec('DaVinci', 'v42r0')
myApp.options = ['code/davinci-grid/ntuple_options_grid.py']
j.application = myApp
```

the second line returns:
```
INFO     Set up App Env at: /afs/cern.ch/user/c/cmarinbe/cmtuser/DaVinciDev_v42r0
```
which seems in line of the folder being created automatically. 

However, when trying to add the input data as we were doing so far:
```
j.inputdata = j.application.readInputData('data/MC_2012_27163003_Beam4000GeV2012MagDownNu2.5Pythia8_Sim08e_Digi13_Trig0x409f0045_Reco14a_Stripping20NoPrescalingFlagged_ALLSTREAMS.DST.py')
```
an error raises:
```
INFO     Started parsing input Data file
ERROR    PythonOptsCmakeParserError:
GangaException:  ("The given directory: '/afs/cern.ch/user/c/cmarinbe/cmtuser/DaVinciDev_v42r0' doesn't exist!",)
ERROR    Error: ApplicationConfigurationError: Unable to parse the job options. Please check options files and extraopts. 
```
I've checked the folder is really not there. I guess it is only created when the job is prepared. I'm not sure why it is needed in order to parse the input data (project dependent?) but we need to figure out what's the best way off doing this now.

cmarinbe:
Ok, my bad, the problem was the project version I was setting. Using: 
```
j = Job()
myApp = prepareGaudiExec('DaVinci','v40r2')
myApp.options = ['code/davinci-grid/ntuple_options_grid.py']
j.application = myApp
j.inputdata = j.application.readInputData('data/MC_2012_27163003_Beam4000GeV2012MagDownNu2.5Pythia8_Sim08e_Digi13_Trig0x409f0045_Reco14a_Stripping20NoPrescalingFlagged_ALLSTREAMS.DST.py')
j.backend = Dirac()
j.submit()
```
[...]

PR #211, opened 2016-11-29
Use GaudiExec in Ganga lessons

cmarinbe:
This should solve #203

### DIFF
--- a/davinci-grid.md
+++ b/davinci-grid.md
@@ -30,24 +30,25 @@ to some special commands provided by `ganga`.
 To create your first `ganga` job type the following:
 
 ```python
-j = Job(application=DaVinci(version='v41r2'))
+j = Job(name='First ganga job')
+myApp = prepareGaudiExec('DaVinci','v41r2')
+myApp.options = ['code/davinci/ntuple_options.py']
+j.application = myApp
+j.application.readInputData('data/MC_2012_27163003_Beam4000GeV2012MagDownNu2.5Pythia8_Sim08e_Digi13_Trig0x409f0045_Reco14a_Stripping20NoPrescalingFlagged_ALLSTREAMS.DST.py')
 j.backend = Dirac()
-j.name = 'First ganga job'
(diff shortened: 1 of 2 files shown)


alexpearce:
LGTM 👍 just in time!

davinci-grid.md
```diff
@@ -30,24 +30,38 @@ to some special commands provided by `ganga`.
 To create your first `ganga` job type the following:
 
 ```python
-j = Job(application=DaVinci(version='v41r2'))
+j = Job(name='First ganga job')
+myApp = prepareGaudiExec('DaVinci','v41r2', myPath='.')
+j.application = myApp
```
rob-c, line 35: Just a daft question but is there some motivation to having the myApp object around after constructing the job?
I keep seeing it in Ganga tutorials but I don't know why people always construct the object then assign it rather than doing this in 1 line.
(If this is historical ignore me :P)

alexpearce: I guess to be explicit about the creation of the object. We could say that a one-liner is possible. 

I guess there's nothing one could do with the GaudiExec object that couldn't be done with j.application?

rmatev: slightly off topic, but IMHO it's worth stressing that `prepareGaudiExec` has observable side effects (does `lb-dev`). Having it on a separate line (even moving it before `Job()` helps with that.

rmatev: Also, if you submit jobs in a loop, you probably want to move it outside.

rob-c: This has been my thoughts, but Ganga explicitly creates a copy of the `myApp` and assigns it to the `job.application` so any further modifications of `myApp` will fail to modify the job application despite the fact you may naively expect that from python.
(What we do is a bit un-pythonic but is required so that we have correct object properties when writing to disk and when copying objects between jobs)

rob-c: I think the code is good and doesn't need changing. I was just wondering if there was something I wasn't aware of in people's workflow as I'd seen almost all tutorials for Ganga done this way.

rmatev: Right. If we want to be more pythonic (i.e. don't suggest that the user can later modify `myApp` and expect a change in the job), we can just construct the job in one go:
```python
j = Job(name='...', application=myApp, ...)
```
I'm not sure though if this is the best for a tutorial.

PR #213, opened 2016-11-30
Add a callout about proxy in eos-storage

renaudin:
I just added a callout to stress out the need of a valid grid proxy to be able to access any grid file using the XRootD protocol.


ibab:
LGTM 🎉

PR #237, opened 2017-04-25
Update how to run ganga

chrisburr:


rmatev:
Maybe worth adding (from Ulrik's announcement, haven't tested it):

To get a specific version, type `ganga --ganga-version 6.5.1`

alexpearce:
There is no `--list` though 😢 Maybe we can add the `ls` we need to find the versions.

chrisburr:
I'd be tempted to exclude this to encourage people just just take the latest version. I don't think there are still good reasons to use old versions?

chrisburr:
For listing versions I'd suggest looking here: https://github.com/ganga-devs/ganga/releases

rmatev:
I have mixed feelings. In the past, more often than not whenever I updated the version I had to spend a lot of time figuring out why something doesn't work, work around it and post bug reports. I don't think it's wasted time as I help improve ganga, but OTOH, I don't want to be forced into it whenever they release the next version.

alexpearce:
Things have gotten better over the past few months, in terms of stability, IMO.

saschastahl:
Yes, but when you are writing your thesis or finishing an analysis, it helps to be able to stay on the same software version.

alexpearce:
At that level you're implicitly dependent on too many other things. For example, if you re-run your Ganga script to make a `Job`, it might be that a site is down that day, or that a file has since been marked `BAD`, so you end up with different results.

I agree in principle, but I don't think it's realistic for Ganga.

PR #238, opened 2017-05-02
Modernise and clarify the lhcb development lesson

lbel:
I've merged the lb-git lesson from second-analysis-steps into this one (that was more convenient than the other way around), and shuffled the content around a bit. It now features some general Cern git steps, lb-git, some notes on documentation, and finally a short section on how to use vanilla git with LHCb software. I plan to teach this lesson at the Impactkit.

### DIFF
--- a/lhcb-dev.md
+++ b/lhcb-dev.md
@@ -1,22 +1,78 @@
 ---
 layout: page
 title: First Steps in LHCb
-subtitle: Developing LHCb Software
-minutes: 10
+subtitle: Using git to develop LHCb Software
+minutes: 45
 ---
 
 > ## Learning Objectives {.objectives}
 > * Learn the basics of how to work with and modify LHCb software packages
+
+> ## Prerequisites {.callout}
+>
+> Before starting, you should have a basic understanding of how to use `git`,
+> similar to what has been
+> [taught](https://lhcb.github.io/analysis-essentials/git/git.html) during the
+> starterkit.
+
+In this lesson, we'll show you a complete workflow for developing the LHCb
+software using the `git` version control system.  At LHCb, we use Gitlab to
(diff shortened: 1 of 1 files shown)


lhcb-dev.md
```diff
@@ -137,23 +208,88 @@ create a new merge request, for example https://gitlab.cern.ch/lhcb/Stripping/me
...
+To get an idea of how a certain component of the LHCb software works, you can
+also access the doxygen documentation. One set of doxygen web pages is
+generated for several related projects, and is linked in all the projects web
+sites, like
+[https://cern.ch/LHCb-release-area/DOC/davinci/](https://cern.ch/LHCb-release-area/DOC/davinci/).
+See also the the [LHCb Computing web page](http://cern.ch/lhcb-comp/) for a
+list of projects.
+
+## Working with a full project checkout
+
```
saschastahl: Here we should point to https://gitlab.cern.ch/lhcb/upgrade-hackathon-setup  or
to @rmatev development to make working with project checkouts more convenient
https://gitlab.cern.ch/lhcb-HLT/trigger-dev.

saschastahl:
Can we point out that the LHCb git commands should only be used for very simple changes, you want to push to the software stack, or when you make some development for yourself or your analysis group only? Having seen many merge requests which went wrong using these commands, I think we should encourage the use of project checkouts + plain git.

lhcb-dev.md
lbel: Good idea! Can both methods be considered "standard" LHCb tools? I could just add both.

lbel:
I'm definitely going to say that very clearly tomorrow, the text could probably be more explicit about it.

lhcb-dev.md
saschastahl: Let's say we would like to make them standard.

PR #240, opened 2017-09-14
Deprecate repository in favour of starterkit-lessons

chrisburr:
We currently have two repositories containing the lessons for the starterkit. This PR removes all the files in this repository in favour of [starterkit-lessons](https://github.com/lhcb/starterkit-lessons) and adds a notice to make people aware of the change. Additionally #241 replaces all of the pages on the site with [redirects](https://stackoverflow.com/a/19717455/2685230).


alexpearce:
My reaction when seeing a diff with nearly 40k deletions:

> "I felt a great disturbance in the Force, as if millions of voices suddenly cried out in terror, and were suddenly silenced. I fear something terrible has happened."

Aren't this PR and #241 in conflict? This one deletes the pages but #241 changes them to have a redirection.

It's 'deprecate' and 'deprecation', rather than 'depreciate' and 'depreciation'.

I was imagining the behaviour being a big banner near the top of every page, and then a 5–10 second timeout before the redirect. I don't care too strongly about the actual implementation though.

chrisburr:
#240 modifies `master`, the #241 modifies the `gh-pages` branch.

The motivation for doing it this way is it hopefully causes google to recognise the redirect quickly and also makes it less likely anyone tries to user or contribute to the old lessons. (Plus it was really easy! 👍)

PR #241, opened 2017-09-14
Replace all pages with redirects to starterkit-lessons (part 2)

chrisburr:
See #240.

PR #242, opened 2017-10-01
Removed SetupProject ganga.

atrisovic:


chrisburr:
Thank you @atrisovic! 🎉 

I've moved this over to lhcb/starterkit-lessons#19 as this repository is being replaced in favour of (it will be clearer once #240 and #241 are merged).

